PR #14, opened 2021-07-07
Move grafana to the operator framework and pebble

rbarry82:
Move to a provider model for the charm, with an example consumer, and clean up the code so we're not directly templating text.

In addition, use Pebble instead of raw pods

### DIFF
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -40,7 +40,7 @@ class GrafanaSourcesChanged(EventBase):
         self.data = snapshot["data"]
 
 
-class GrafanaSourceEvents(CharmEvents):
+class GrafanaSourceEvents(ObjectEvents):
     """Events raised by :class:`GrafanaSourceEvents`"""
 
     # We are emitting multiple events for the same thing due to the way Grafana provisions
@@ -223,6 +223,8 @@ class GrafanaSourceProvider(ProviderBase):
         """
         super().__init__(charm, name, service, version)
         self.charm = charm
+        self.charm.unit.prov_network = self.model.get_binding(self.name).network.bind_address
+        self.charm.app.prov_network = self.model.get_binding(self.name).network.bind_address
         events = self.charm.on[name]
 
         self._stored.set_default(sources=dict())  # available data sources
@@ -247,6 +249,13 @@ class GrafanaSourceProvider(ProviderBase):
         """
         if not self.charm.unit.is_leader():
             return
+
+        logger.info("RELATION APP: {}".format(event.relation.app))
+        logger.info("RELATION APP: {}".format(dir(event.relation.app)))
+
+        for u in event.relation.units:
+            logger.info("UNIT: {}".format(u))
+            logger.info("UNIT: {}".format(dir(u)))
 
         rel = event.relation
         data = (
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -3,12 +3,11 @@ import logging
 
 from ops.charm import (
     CharmBase,
-    CharmEvents,
     RelationBrokenEvent,
     RelationChangedEvent,
     RelationJoinedEvent,
 )
-from ops.framework import EventBase, EventSource, StoredState
+from ops.framework import EventBase, EventSource, ObjectEvents, StoredState
 from ops.relation import ConsumerBase, ProviderBase
 from typing import Dict, List, Optional
 
@@ -223,8 +222,6 @@ class GrafanaSourceProvider(ProviderBase):
         """
         super().__init__(charm, name, service, version)
         self.charm = charm
-        self.charm.unit.prov_network = self.model.get_binding(self.name).network.bind_address
-        self.charm.app.prov_network = self.model.get_binding(self.name).network.bind_address
         events = self.charm.on[name]
 
         self._stored.set_default(sources=dict())  # available data sources
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -1,0 +1,336 @@
+import json
+import logging
+
+from ops.charm import (
+    CharmBase,
+    CharmEvents,
+    RelationBrokenEvent,
+    RelationChangedEvent,
+    RelationJoinedEvent,
+)
+from ops.framework import EventBase, EventSource, StoredState
+from ops.relation import ConsumerBase, ProviderBase
+from typing import Dict, List, Optional
+
+
+LIBID = "987654321"
+LIBAPI = 1
+LIBPATCH = 0
+
+logger = logging.getLogger(__name__)
+
+
+class SourceFieldsMissingError(Exception):
+    pass
+
+
+class GrafanaSourcesChanged(EventBase):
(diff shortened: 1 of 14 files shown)


rbarry82:
More PR splits

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+        self._stored.set_default(sources_to_delete=set())
+
+        self.framework.observe(
+            events.relation_changed, self.on_grafana_source_relation_changed
+        )
+        self.framework.observe(
+            events.relation_broken, self.on_grafana_source_relation_broken
+        )
+
+    def on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> None:
```
sed-i, line 238: leading underscore missing?

metadata.yaml
```diff
@@ -1,38 +1,41 @@
...
+containers:
+    grafana:
+        resource: grafana-image
+        mounts:
+            - storage: database
+              location: /var/lib/grafana
 provides:
     grafana-source:
-        interface: grafana-datasource
+        interface: grafana_datasource
```
sed-i, line 24: just curious why the interface name is not identical to the relation name

rbarry82: The interface name is separate from the actual relation (see also Prometheus as `monitoring` vs `prometheus_scrape`). A hyphen in the interface name more closely matches the natural output in `juju status`, such as `grafana-k8s:grafana-source grafana_datasource` instead of `grafana-k8s:grafana_datasource ...`

balbirthomas (approved):
Very clean. Simple and easy to read. Awesome. Impressive.

src/charm.py
```diff
@@ -1,60 +1,67 @@
 #! /usr/bin/env python3
 # -*- coding: utf-8 -*-
```
sed-i, line 2: not needed - this is the default for python3

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
         super().__init__(*args)
 
-        # -- get image information
-        self.image = OCIImageResource(self, 'grafana-image')
-
-        # -- standard hooks
-        self.framework.observe(self.on.config_changed, self.on_config_changed)
-        self.framework.observe(self.on.update_status, self.on_update_status)
+        # -- initialize states --
+        self.name = "grafana"
```
sed-i, line 80: why not class attribute?

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
+        self.container = self.unit.get_container(self.name)
+        self.grafana_service = Grafana("localhost", self.model.config["port"])
+        self.grafana_config_ini_hash = None
+        self.grafana_datasources_hash = None
```
sed-i, line 84: Shouldn't these be in `_stored`?

rbarry82: There is no reason to store these in the model, since they are only used for lookups inside running execution

src/charm.py
rbarry82: This also hints for editors

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
+        When a grafana-source is added or modified, update the config
 
-    @property
-    def has_db(self) -> bool:
-        """Only consider a DB connection if we have config info."""
-        return len(self.datastore.database) > 0
+        Args:
+            event: a :class:`GrafanaSourceEvents` instance sent from the consumer
+        """
+        self._configure(event)
```
sed-i, line 133: not sure why you pass the event to _configure, which shouldn't (and in fact doesn't) need it

src/charm.py
rbarry82: Simply shorter

src/charm.py
rbarry82: In the case that `_configure` should respond to other events (or ignore them if it's shutting down so I don't need the `pebble.ConnectionError` handling), it's easier to change it in one place (the signature for `_configure`) than a whole bunch of them

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
-        if len(missing_fields) > 0:
-            log.error("Missing required data fields for grafana-source "
-                      "relation: {}".format(missing_fields))
-            self._remove_source_from_datastore(event.relation.id)
-            return
+    @property
+    def has_peers(self) -> bool:
+        """Check whether or nto there are any other Grafanas as peers"""
+        rel = self.model.get_relation(PEER)
+        return len(rel.units) > 0 if rel is not None else False
```
sed-i, line 214: Consider:
```python
return rel is not None and len(rel.units) > 0
```

rbarry82: While shorter, this is the closest Python equivalent to a traditional ternary. See PEP308.

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
+                return False
 
-        # in the case where we have peers but no DB connection,
-        # don't set the pod spec until it is resolved
-        if self.unit.status == BlockedStatus('Need database relation for HA.'):
-            log.error('Application is in a blocked state. '
-                      'Please resolve before pod spec can be set.')
-            return
+        if not self._stored.pebble_ready:
+            self.unit.status = WaitingStatus("Waiting for Pebble startup to complete")
```
sed-i, line 451: Shouldn't this be `MaintenanceStatus`?

src/charm.py
```diff
@@ -64,439 +71,439 @@ class GrafanaK8s(CharmBase):
...
 
-        if not self.unit.is_leader():
-            self.unit.status = ActiveStatus()
-            return
+        layer = self._build_layer()
+        if self.has_peers and not layer.services.grafana.environment.GF_DATABASE_URL:
+            self.unit.status = MaintenanceStatus("Related MySQL not yet ready")
```
sed-i, line 455: Shouldn't this be `WaitingStatus`?

sed-i (dismissed):

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+    def snapshot(self) -> Dict:
+        """Save grafana source information"""
+        return {"data": self.data}
+
+    def restore(self, snapshot) -> None:
+        """Restore grafana source information"""
+        self.data = snapshot["data"]
+
+
+class GrafanaSourceEvents(CharmEvents):
```
sed-i, line 43: `ObjectEvents`?

sed-i (approved):

requirements.txt
```diff
@@ -1,2 +1,4 @@
-ops
-git+https://github.com/juju-solutions/resource-oci-image/@c5778285d332edf3d9a538f9d0c06154b7ec1b0b#egg=oci-image
\ No newline at end of file
+git+https://github.com/canonical/operator/#egg=ops
+semantic_version
```
sed-i, line 2: not needed anymore :D https://github.com/canonical/operator/pull/564

dstathis:
You committed an __init__.py. These are not necessary for charms

requirements.txt
rbarry82: Not merged yet

sed-i (approved):

balbirthomas (approved):

PR #17, opened 2021-07-23
Juju topology

rbarry82:
Move to Juju topology to match the new paradigm

### DIFF
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -124,6 +124,9 @@ class GrafanaSourceConsumer(ConsumerBase):
         On a relation_joined event, inform the provider about the source
         configuration
         """
+        if not self.charm.unit.is_leader():
+            return
+
         event.relation.data[self.charm.unit]["grafana_source_host"] = "{}:{}".format(
             str(self.charm.model.get_binding(event.relation).network.bind_address),
             self._source_port,
@@ -218,6 +221,9 @@ class GrafanaSourceProvider(ProviderBase):
         The Grafana charm can then respond to the event to update its
         configuration.
         """
+        if not self.charm.unit.is_leader():
+            return
+
         sources = {}
 
         for rel in self.charm.model.relations[self.name]:
@@ -235,19 +241,18 @@ class GrafanaSourceProvider(ProviderBase):
         consumers which we can pass back to the charm
         """
 
-        source_data = json.loads(rel.data[rel.app].get("grafana_source_data", ""))
+        source_data = json.loads(rel.data[rel.app].get("grafana_source_data", "{}"))
         if not source_data:
             return
 
         data = []
 
         for unit_name, host_addr in self._relation_hosts(rel).items():
-            unique_source_name = "juju_{}_{}_{}_{}_{}".format(
+            unique_source_name = "juju_{}_{}_{}_{}".format(
                 source_data["model"],
                 source_data["model_uuid"],
                 source_data["application"],
                 unit_name.split("/")[1],
-                rel.id,
             )
 
             host_data = {
@@ -257,7 +262,6 @@ class GrafanaSourceProvider(ProviderBase):
             }
 
             data.append(host_data)
-
         return data
 
     def _relation_hosts(self, rel: Relation) -> Dict:
@@ -297,7 +301,6 @@ class GrafanaSourceProvider(ProviderBase):
         name to the list of sources to remove when a relation is
         broken.
         """
-
         logger.debug("Removing all data for relation: {}".format(rel_id))
 
         try:
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -124,13 +124,7 @@ class GrafanaSourceConsumer(ConsumerBase):
         On a relation_joined event, inform the provider about the source
         configuration
         """
-        if not self.charm.unit.is_leader():
-            return
-
-        event.relation.data[self.charm.unit]["grafana_source_host"] = "{}:{}".format(
-            str(self.charm.model.get_binding(event.relation).network.bind_address),
-            self._source_port,
-        )
+        self._set_unit_ip(event)
 
         if not self.charm.unit.is_leader():
             return
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -297,14 +297,11 @@ class GrafanaSourceProvider(ProviderBase):
         """
         logger.debug("Removing all data for relation: {}".format(rel_id))
 
-        try:
-            removed_source = self._stored.sources.pop(rel_id, None)
-            if removed_source:
-                for host in removed_source:
-                    self._remove_source(host["source-name"])
-                self.on.sources_to_delete_changed.emit()
-        except KeyError:
-            logger.warning("Could not remove source for relation: {}".format(rel_id))
+        removed_source = self._stored.sources.pop(rel_id, None)
+        if removed_source:
+            for host in removed_source:
+                self._remove_source(host["source-name"])
+            self.on.sources_to_delete_changed.emit()
 
     def _remove_source(self, source_name: str) -> None:
         """Remove a datasource by name"""
--- a/lib/charms/grafana_k8s/v1/grafana_source.py
+++ b/lib/charms/grafana_k8s/v1/grafana_source.py
@@ -3,11 +3,13 @@ import logging
 
 from ops.charm import (
     CharmBase,
+    CharmEvents,
     RelationBrokenEvent,
     RelationChangedEvent,
     RelationJoinedEvent,
 )
 from ops.framework import EventBase, EventSource, ObjectEvents, StoredState
+from ops.model import Relation
 from ops.relation import ConsumerBase, ProviderBase
 from typing import Dict, List, Optional
 
@@ -55,7 +57,14 @@ class GrafanaSourceConsumer(ConsumerBase):
     _stored = StoredState()
 
     def __init__(
-        self, charm: CharmBase, name: str, consumes: dict, multi=False
+        self,
+        charm: CharmBase,
+        name: str,
+        consumes: dict,
+        refresh_event: CharmEvents,
+        source_type: Optional[str] = "prometheus",
+        source_port: Optional[str] = "9090",
+        multi: Optional[bool] = False,
     ) -> None:
         """Construct a Grafana charm client.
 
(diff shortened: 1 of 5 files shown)


balbirthomas (approved):

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -55,7 +57,14 @@ class GrafanaSourceConsumer(ConsumerBase):
     _stored = StoredState()
 
     def __init__(
-        self, charm: CharmBase, name: str, consumes: dict, multi=False
+        self,
+        charm: CharmBase,
+        name: str,
+        consumes: dict,
+        refresh_event: CharmEvents,
+        source_type: Optional[str] = "prometheus",
```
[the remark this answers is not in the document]
rbarry82: Reasonable defaults. For now, we can safely assume that most, if not all sources will be prometheus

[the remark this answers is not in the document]
mmanciop: That now may change in a few weeks as we look into plotting Alertmanager alerts and Loki logs. But not a blocker for this PR by any stretch of the imagination.

mmanciop (dismissed):

lib/charms/grafana_k8s/v1/grafana_source.py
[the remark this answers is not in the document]
rbarry82: On the plus side, they're just defaults. They can easily be overridden by consumers, and AlertManager/Loki will hopefully use the library anyway, so it's a different argument in the constructor

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -88,122 +97,85 @@ def __init__(
                 this is `grafana-source`. The values of the dictionary
                 are corresponding minimal acceptable semantic versions
                 for the service.
+            refresh_event: a :class:`CharmEvents` event on which the IP
+                address should be refreshed in case of pod or
+                machine/VM restart.
+            source_type an optional (default `prometheus`) source type
+                required for Grafana configuration
+            source_port an optional (default `9090`) source port
+                required for Grafana configuration
```
sed-i, line 106: After reading the docstrings, the usage/purpose/motivation of the new params are still unclear to me.

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -88,122 +97,85 @@ def __init__(
...
-                `multi` mode.
+        On a relation_joined event, inform the provider about the source
+        configuration
         """
-        rel = self.framework.model.get_relation(self.name, rel_id)
-
-        if rel_id is None:
-            rel_id = rel.id
+        if not self.charm.unit.is_leader():
+            return
```
sed-i, line 128: can probably remove the guard here (there's another one below before app data)

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -88,122 +97,85 @@ def __init__(
...
 
-    def _source_name_in_use(self, source_name) -> bool:
-        return any(
-            [s["source-name"] == source_name for s in self._stored.sources.values()]
-        )
+        Each time a consumer charm container is restarted it updates its own
+        host address in the unit relation data for the Prometheus provider.
+        """
+        for relation in self.charm.model.relations[self.name]:
+            relation.data[self.charm.unit]["grafana_source_host"] = "{}:{}".format(
```
sed-i, line 166: seems like exact duplicate from line 130

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -241,42 +215,72 @@ def _on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> No
...
-        is available in the app's datastore object. The Grafana charm can
-        then respond to the event to update its configuration
+        is available in the app's datastore object. This data is set using
+        Juju application topology.
+
+        The Grafana charm can then respond to the event to update its
+        configuration.
         """
         if not self.charm.unit.is_leader():
             return
```
sed-i, line 225: still needed?

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -241,42 +215,72 @@ def _on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> No
...
+        data = []
 
-        data["isDefault"] = (
-            "false" if dict(self._stored.sources) else "true"
-        )
+        for unit_name, host_addr in self._relation_hosts(rel).items():
+            unique_source_name = "juju_{}_{}_{}_{}".format(
+                source_data["model"],
+                source_data["model_uuid"],
+                source_data["application"],
```
sed-i, line 254: Potential `KeyError`. Just saying.

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -241,42 +215,72 @@ def _on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> No
...
+                unit_name.split("/")[1],
```
sed-i, line 255:
```suggestion
                unit_name.rsplit("/", 1)[1],
```

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -241,42 +215,72 @@ def _on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> No
...
+            rel: An `ops.model.Relation` object for which the host name to
+                address mapping is required.
+        Returns:
+            A dictionary that maps unit names to unit addresses for
+            the specified relation.
+        """
+        hosts = {}
+        for unit in rel.units:
+            host_address = rel.data[unit].get("grafana_source_host")
+            if not host_address:
```
sed-i, line 280: `KeyError` or is it `defaultdict`?

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: For when we add other consumers (Loki, Alertmanager, whatever). They will require different types

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: No, this was part of a messy merge

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -241,42 +215,72 @@ def _on_grafana_source_relation_changed(self, event: RelationChangedEvent) -> No
...
+                continue
```
sed-i, line 281: side note: we should probably have some facility in OF for filtering out provider units or consumer units

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Yes. It guards here instead of at `_get_source_config`

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: These are a mandatory part of the interface which are set by the consumer.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: We do. It's `ops.relation` and the relationships they establish there, in these libraries

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -297,23 +301,27 @@ def _remove_source_from_datastore(self, rel_id: int) -> None:
...
         logger.debug("Removing all data for relation: {}".format(rel_id))
 
         try:
             removed_source = self._stored.sources.pop(rel_id, None)
             if removed_source:
-                self._stored.sources_to_delete.add(removed_source["source-name"])
+                for host in removed_source:
+                    self._remove_source(host["source-name"])
                 self.on.sources_to_delete_changed.emit()
         except KeyError:
```
sed-i, line 306: this means that `sources_to_delete_changed` will not be emitted even if the first N removals succeeded.
not sure if that is intended.

rbarry82: Since we're looping over the same object, it is not possible for the first N to success and N+1 to fail. It's a single relation ID per invocation of the function. If we cannot pop one, the entire thing is a `KeyError`. if we can, we don't need to emit the event N times.

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -297,23 +301,27 @@ def _remove_source_from_datastore(self, rel_id: int) -> None:
...
             logger.warning("Could not remove source for relation: {}".format(rel_id))
 
+    def _remove_source(self, source_name: str) -> None:
+        """Remove a datasource by name"""
+        self._stored.sources_to_delete.add(source_name)
```
sed-i, line 311: could you please document how `_stored.sources_to_delete` (persistent private variable) and `sources_to_delete_changed` (even observed by the provider charm) interplay?

sed-i (approved):
LGTM. BTW, what do you have against the walrus?

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: what are they used for?
When writing a charm, if i check the doc and see "source type required for X", I have no idea what to feed there.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: There's not much to document, or not anything more than `sources_changed`.

Grafana [requires a separate list](https://grafana.com/docs/grafana/latest/administration/provisioning/) of datasources which have been removed. If they are not in `deleteDatasources` in the configuration, it doesn't invalidate them from the list.

This is very obvious from the charm code, where it's [only used in a single place](https://github.com/canonical/grafana-operator/blob/28997f3791b071d358defcf39b1e47ea0e8abb7b/src/charm.py#L501), and is not specific to the Charm operation.

I'm against documenting this for the same reason I'd be against documenting the complete array of configuration available in Prometheus, Alertmanager, Loki, or anything else.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Presumably, if you are a charm author who is using the Grafana consumer library and you want to set a non-default value, you know enough to be able to [check the docs](https://grafana.com/docs/grafana/latest/administration/provisioning/).

Even the official docs do not document this:
```
    # <string, required> datasource type. Required
    type: graphite
```

We are not going to either. I couldn't even provide a link to their docs if they had one, because this hint is about as good as their docs

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: I thought the `except KeyError` is for `host["source-name"]`. Your pop has a default so there won't be any KeyError.
What am I missing here?

rbarry82: It's from a much older version of the charm, and the logic around it hasn't been changed. The chances of a value making it *into* `self._stored.sources` without having a `source-name` key are zero, so there will never be a `KeyError` here. I'll cut the dead code, which I missed on reviewing it, because this part of the library never gets touched

rbarry82:
The support matrix for the Operator Framework goes back to Python 3.5 (which should be removed soon, so let's say 3.6). Keeping backwards compatibility is important.

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: no worries, just a confusion on my part because `_remove_source` doesn't remove anything.

sed-i:
> The support matrix for the Operator Framework goes back to Python 3.5 (which should be removed soon, so let's say 3.6). Keeping backwards compatibility is important.

This is a k8s charm, so py3.8 iiuc.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Grafana hasn't changed this since `5.x`. They are on `8.x` now. I sort of think they don't see it as a priority.

Their docs for templates are also very bad. The guidelines seem to be "build the dashboard in the UI, then export it". There's almost no response on requests for "provide a way to reload the configuration without restarting" (this is technically supported through an API endpoint only, and only with basic auth, and is also on a random forum post and not the official docs).

Trying to document design decisions in the charm when the actual product can't be bothered to document anything is rough

rbarry82:
Sure, but it's also habits, style, and consistency. If I habitually use new features, code I write for other projects which have a testsuite that runs against older versions will fail. It's mostly mental hygiene.

sed-i (approved):

sed-i (approved):

balbirthomas (approved):

PR #20, opened 2021-07-27
Add a dashboard library

rbarry82:
Add a library to provide/consumer dashboard data, and check the
validity of the relations in the set to set appropriate messages.

Add tests to match


sed-i (dismissed):

balbirthomas (approved):

rbarry82:
Updated `CONTRIBUTING.md` also

src/charm.py
```diff
@@ -130,6 +140,7 @@ def _on_grafana_source_changed(self, event: GrafanaSourceEvents) -> None:
         Args:
             event: a :class:`GrafanaSourceEvents` instance sent from the consumer
         """
+        self.dashboard_provider.renew_dashboards(self.source_provider.sources)
```
sed-i, line 143: should this also go into on_update_status?

src/charm.py
```diff
@@ -148,33 +159,33 @@ def _configure(self, _) -> None:
 
         # Generate a new base config and see if it differs from what we have.
         # If it does, store it and signal that we should restart Grafana
-        grafana_config_ini = self.generate_grafana_config()
+        grafana_config_ini = self._generate_grafana_config()
         config_ini_hash = hashlib.sha256(
             str(grafana_config_ini).encode("utf-8")
         ).hexdigest()
```
sed-i, line 165: code duplication. def sha256?

src/charm.py
```diff
@@ -148,33 +159,33 @@ def _configure(self, _) -> None:
...
         if not self.grafana_config_ini_hash == config_ini_hash:
```
sed-i, line 166: seems like a job for _stored, because on_stop will happen after reinstantiation of the charm

src/charm.py
```diff
@@ -148,33 +159,33 @@ def _configure(self, _) -> None:
...
+            self._update_datasource_config(grafana_datasources)
+            logger.info("Pushed new grafana-k8s datasource configuration")
 
             restart = True
 
         if restart:
             self.restart_grafana()
 
-    def update_datasource_config(self, config: str) -> None:
+    def _update_datasource_config(self, config: str) -> None:
```
sed-i, line 188: consider renaming to push iv favour of consistent terminology

sed-i (dismissed):

src/charm.py
rbarry82: Only matters if Grafana source relations change

src/charm.py
rbarry82: Yes, but in that case, the config will be identical and will never be pushed. The point is that this is dynamic if there are new DB relations.

sed-i (approved):

src/charm.py
```diff
@@ -147,12 +148,15 @@ def _on_stop(self, _) -> None:
...
-    def _configure(self, _) -> None:
+    def _configure(self, event: CharmEvents) -> None:
         """
         Generate configuration files and check the sums against what is
         already stored in the charm. If either the base Grafana config
         or the datasource config differs, restart Grafana.
         """
+        if not self._stored.pebble_ready:
+            logger.warning("Pebble is not ready yet. Deferring event.")
+            event.defer()
```
sed-i, line 159: at first glance, it makes sense no to return after defer.
at second glace - this feels a bit convoluted.
Greylog's pattern of `return False` etc. doesn't work for you here?

src/charm.py
rbarry82: The docstring and unit test for a sha256 function would be longer than the duplication, and force reviewers to check another method which does has only a single line.

sed-i (dismissed):

PR #24, opened 2021-09-08
Add pyproject.toml and update tox.ini

jnsgruk:
As the description says, introduce a `tox.ini` and a `pyproject.toml` to the project.

Note: linting fails at this stage, fixes needed for it to pass are in #25


rbarry82 (approved):
LGTM after an update to make mypy happy.

It definitely doesn't like `lib/charms/...` compared to `import charms.` in `src/charm.py`, though. We should either get :star:craft to redefine the build/deployment strategy, or run mypy once against `tests/` and once against everything else.

I'm surprised this hasn't cropped up in other charms.

jnsgruk:
> LGTM after an update to make mypy happy.
> 
> It definitely doesn't like `lib/charms/...` compared to `import charms.` in `src/charm.py`, though. We should either get craft to redefine the build/deployment strategy, or run mypy once against `tests/` and once against everything else.
> 
> I'm surprised this hasn't cropped up in other charms.

Fixed ;-)

It's because you were importing libraries as `lib.charms.whatever`, you can just import them as `charms.` because of the modified `PYTHONPATH`. With the last commit, the following all pass:

```
tox -e lint
tox -e static
tox -e unit
```

rbarry82 (approved):

tox.ini
```diff
@@ -0,0 +1,79 @@
...
+# See LICENSE file for licensing details.
+[tox]
+skipsdist=True
+skip_missing_interpreters = True
+envlist = lint, static, unit
+
+[vars]
+src_path = {toxinidir}/src/
+tst_path = {toxinidir}/tests/
+lib_path = {toxinidir}/lib/charms/grafana_k8s/
```
sed-i, line 11: the `/` suffix is redundant?

sed-i (approved):

jnsgruk:
Merge whenever you're ready @rbarry82 :)

PR #26, opened 2021-09-09
Remove ops.relation, swap requires and provides in metadata

rbarry82:
Dylan's intuition was correct -- dumping `ops.relation` from the tree, changing the parent to `Object`, and removing the args from the old parent class works as expected.

Only used `.ready()` in one place in the charm

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -93,7 +93,7 @@ class GrafanaCharm(CharmBase):
         self.framework.observe(self.on.stop, self._on_stop)
 
         # -- grafana_source relation observations
-        self.source_provider = GrafanaSourceProvider(self, "grafana-source", "grafana", VERSION)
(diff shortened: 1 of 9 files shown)


balbirthomas (dismissed):

src/charm.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, *args):
         self.framework.observe(self.on.stop, self._on_stop)
 
         # -- grafana_source relation observations
-        self.source_provider = GrafanaSourceProvider(self, "grafana-source", "grafana", VERSION)
```
sed-i, line 96: VERSION is now unused. Remove?

sed-i (approved):

sed-i (approved):

PR #29, opened 2021-09-22
Fix cross-model dashboards

rbarry82:

PR #30, opened 2021-09-24
Prevent breaking Jinja2 3x changes

mmanciop:
Due to https://github.com/pallets/jinja/issues/1496, we cannot use the latest Jinja2 packages inside the operator. The breaking change was introduced in Jinja2's 3x branch, and this change pins the `jinja2` dependency to the latest 2x release.


rbarry82 (approved):

rbarry82:
Already integrated into #29

PR #37, opened 2021-10-11
Make sure Grafana has been initialized before dashboards are added

rbarry82:


rbarry82:
Extra argument is removed.

After a goose chase, @mmanciop's new "dashboards not visible" issue was due to an overly aggressive replacement on the provider/consumer library flip. Grafana logs, even in debug, were not useful, of course.

`providers` [is a fixed key](https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards) in the provisioning discovery YAML which cannot be replaced with "consumers" and have it work.

jnsgruk (approved):

PR #38, opened 2021-10-12
Fix typing issue with relations, consistently use Relation as argument

mmanciop:
Fix issue on `relation_broken` in which we treat a `relation_id` as a `relation`, consistently use `Relation` as argument type for private methods rather than `relation_id`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -348,7 +348,7 @@ class GrafanaCharm(CharmBase):
         ]
         if len(missing_fields) > 0:
             raise SourceFieldsMissingError(
-                "Missing required data fields for database relation: {}".format(missing_fields)
+                f"Missing required data fields for database relation: {missing_fields}"
             )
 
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):

src/charm.py
```diff
@@ -348,8 +348,7 @@ def _on_database_changed(self, event: RelationChangedEvent) -> None:
         ]
         if len(missing_fields) > 0:
             raise SourceFieldsMissingError(
-                "Missing required data fields for grafana-k8s database "
-                "relation: {}".format(missing_fields)
+                "Missing required data fields for database relation: {}".format(missing_fields)
```
jnsgruk, line 351:
```suggestion
                f"Missing required data fields for database relation: {missing_fields}"
```

jnsgruk (approved):

PR #40, opened 2021-10-18
Use LZMA instead of zlib

rbarry82:
LZMA is about 10% more efficient in the worst case, and only increases as complexity goes up

### DIFF
--- a/tests/test_dashboard_provider.py
+++ b/tests/test_dashboard_provider.py
@@ -63,7 +63,7 @@ class ProviderCharm(CharmBase):
         self.framework.observe(
             self.provider.on.dashboard_status_changed,
             self._on_dashboard_status_changed,
-        )
+        )  # type: ignore[arg-type]
 
     def _on_dashboard_status_changed(self, event):
         if event.valid:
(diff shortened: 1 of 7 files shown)


tests/test_dashboard_provider.py
```diff
@@ -19,7 +19,7 @@
 RELATION_TEMPLATES_DATA = {
     "file:first": {
         "charm": "provider-tester",
-        "content": "eNorSS0uiU/LLCou4QIAG8EEUg==",
+        "content": "/Td6WFoAAATm1rRGAgAhARYAAAB0L+WjAQAKdGVzdF9maXJzdAoAAIC4BxCQe2GHAAEjC8Ib/QkftvN9AQAAAAAEWVo=",
```
jameinel, line 22: Interesting that we did this to 'make it more efficient' but in the case of short strings, it is clearly much much less :)

jameinel:
(forget my earlier XZ statement. I do believe there was an issue around it,
but it seems my memory for what is recommended was incorrect. Also Python
uses lzma to do .xz decoding.)

tests/test_dashboard_provider.py
Abuelodelanada: This is true, but we are compressing grafana dashboards that are kilometric json objects. So if I understand correctly most of the time we are going to see the profit.

Abuelodelanada (approved):
Looks good to me

Abuelodelanada (approved):
LGTM

simskij (approved):

PR #41, opened 2021-10-29
Allow retrieving the default admin password with actions

rbarry82:


rbarry82:
Rebased on top of the lzma patch...

actions.yaml
```diff
@@ -0,0 +1,2 @@
+get-admin-password:
```
mmanciop, line 1: Retrieve the current password for the “admin” account in Grafana

mmanciop (changes requested):
Needs some docu about the way the password works in the README and improving the description in actions.yaml then can be merged

src/charm.py
```diff
@@ -529,6 +536,25 @@ def _generate_datasource_config(self) -> str:
...
+            self._stored.admin_password = self._generate_password()
+
+        return self._stored.admin_password
+
+    def _generate_password(self) -> str:
+        """Generates a random 12 character password."""
+        # Really limited by what can be passed into shell commands, since this all goes
+        # through subprocess. So much for complex password
+        chars = string.ascii_letters + string.digits
+        return "".join(secrets.choice(chars) for _ in range(12))
```
simskij, line 556: I get the point about not using non ascii/digit letters, but why limit ourselves to twelve positions? Seems like we could use way longer default passwords to increase the entropy greatly.

actions.yaml
simskij: @mmanciop I'm a little bit opposed to this from a security standpoint. It should only get the *initial* password, right? If the user changes the password, the expected behavior, from my point of view, would be that it becomes non-retrievable (but maybe resettable through an action).

jnsgruk: Agreed

src/charm.py
rbarry82: Basically because it exceeds the baseline for PCI-DSS, ISO2700[12], and NIST 800-63. 

We could certainly add more characters, but considering that a change is required anyway, meeting baseline security requirements for every major framework is hopefully sufficient.

actions.yaml
rbarry82: Do we want to make it resettable via an action? This can be done with `grafana-cli` in the workload container, but I'm not knowledgeable enough about Juju's RBAC model to guess whether letting Juju admins reset admin passwords for deployed workloads iss sane or not.

mmanciop:
@simskij i think having for now the admin password generated by The charmed operator, retrievable via action and not yet modifiable via juju apis to be a good enough stopgap until Juju secrets land.

simskij:
> @simskij i think having for now the admin password generated by The charmed operator, retrievable via action and not yet modifiable via juju apis to be a good enough stopgap until Juju secrets land.

I agree! My main point is that user-changed passwords never should be retrievable.

rbarry82:
@simskij @mmanciop if we're agreed, let's see some :+1:s!

mmanciop (approved):

simskij (approved):

mmanciop:
@rbarry82 what is preventing us from squashing, merging and publishing a new `edge`?

rbarry82:
Only that I missed the github notification about the second approval

mmanciop:
@rbarry82 I really wish we would use more the "Squash and merge". What is the value to have a commit with message "Linters" in the `main` branch? :-(

rbarry82:
I really don't have a strong preference on it, but there's no shortage of opinions titled "Squash merge considered harmful" ;)

mmanciop:
If the overall PR changeset is huge, I even agree. But then, if the sequence of commits is an intended artifact we wish to preserve, you gotta force-push edits to the PR where the linting is applied where it matters and I shall not complain :D

PR #42, opened 2021-11-18
3.5 ify

rbarry82:
No f-strings, no inline type annotations

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -2,22 +2,26 @@ name: Run-Tests
 on: [pull_request]
 
 jobs:
-  static-analysis:
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        python-version: [3.5, 3.8]
-
-    name: Static analysis
+  lib-check:
+    name: Static analysis of /lib for Python 3.5
+    runs-on: ubuntu-latest
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.5
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.5
+    - name: Install dependencies
+      run: python3 -m pip install tox
+    - name: Run static analysis for /lib for 3.5
+      run: tox -vve static-libs-3.5
+  static:
+    name: Static analysis
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
     - name: Install dependencies
       run: python3 -m pip install tox
     - name: Run static analysis
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -24,20 +24,10 @@ jobs:
       run: tox -vve static
   lint:
     name: Lint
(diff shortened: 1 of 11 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -178,7 +178,7 @@ def _validate_relation_by_interface_and_direction(
     if relation_name not in charm.meta.relations:
         raise RelationNotFoundError(relation_name)
 
-    relation: RelationMeta = charm.meta.relations[relation_name]
+    relation = charm.meta.relations[relation_name]
```
sed-i, line 181: Could still use a `# type:` comment :)

sed-i (approved):

sed-i (approved):

sed-i (approved):

.github/workflows/run_tests.yaml
```diff
@@ -3,11 +3,21 @@ on: [pull_request]
...
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: python3 -m pip install tox
     - name: Run static analysis
```
sed-i, line 23: Just realized that this change as-is is checking everything for 3.5, not just the lib.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -97,9 +98,8 @@ def __init__(
         self._relation_name = relation_name
         self.expected_relation_interface = expected_relation_role
         self.actual_relation_role = actual_relation_role
-        self.message = (
-            f"The '{relation_name}' relation has role '{repr(actual_relation_role)}' "
-            f"rather than the expected '{repr(expected_relation_role)}'"
+        self.message = "The '{}' relation has role '{}' " "rather than the expected '{}'".format(
```
simskij, line 101:
```suggestion
        self.message = "The '{}' relation has role '{}' rather than the expected '{}'".format(
```
if fmt is gonna insist on it being on the same line, there's no need to split the string 😄

simskij (dismissed):

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: Bad `black` :/

sed-i (approved):

Abuelodelanada (approved):

.github/workflows/run_tests.yaml
```diff
@@ -2,7 +2,21 @@ name: Run-Tests
...
+  lib-check:
+    name: Static analysis of /lib for Python 3.5
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up Python 3.5
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.5
```
dstathis, line 14: Not a suggestion, just a question:

Are you compiling python 3.5 in order to run this manually?

dstathis (approved):

.github/workflows/run_tests.yaml
rbarry82: AFAIK, it grabs a version, puts it in `/opt`, then puts that version on the front of `$PATH`, so it's a fetch, but not compile

PR #44, opened 2021-11-24
Add an aggregator for machine charms. Use more unique IDs which strip out the LZMA signatures

rbarry82:
Backend library requirements for the changes made to LMA proxy.


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            "content": content,
+            "juju_topology": self._juju_topology(event),
+        }
+
+    def _juju_topology(self, event: RelationEvent) -> Dict:
+        return {
+            "model": self._charm.model.name,
+            "model_uuid": self._charm.model.uuid,
+            "application": event.app.name,
+            "unit": event.unit.name,
```
sed-i, line 937: [afaik](https://github.com/canonical/prometheus-operator/blob/5ffc1016ac7d2f56c7d862b5e81e33dc05c99234/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L528), "unit" is not a valid JT key.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            logger.warning(
+                "Invalid Grafana dashboards folder at %s: %s",
+                e.grafana_dashboards_absolute_path,
+                e.message,
+            )
+
+        if dashboards_path:
+            for path in filter(Path.is_file, Path(dashboards_path).glob("*.tmpl")):
+                if event.app.name in path.name:
+                    id = f"file:{path.stem}"
```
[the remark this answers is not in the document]
simskij:
```suggestion
                    id = "file:{}".format(path.stem)
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
simskij: [it is](https://juju.is/blog/model-driven-observability-part-2-juju-topology-metrics) 😄 

> ### Unit identifier
> When you deploy a charm, you can effortlessly scale it up and down. Each instance is called a unit. One of the many design decisions of Juju that I love (more on this later) is that units have a fixed, predictable identity: when you scale a Juju application to, say, three units, each instance has a stable identifier built on the Juju application name and an ordinal number starting from zero, for example “users-db/0”, “users-db/1” and “users-db/2”. When one of the units is restarted, because its charm is updated, or it crashes, a new unit with the same identifier takes its place! This also has implications when scaling down a Juju application: when scaling down the users-db application from three units to two, you know that the “users-db/2” application is getting the boot. (By the way, the predictability in terms of which unit is scaled down is a very nice property of Juju in terms of software operations.)

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+
+        for grafana_relation in self.model.relations[self._grafana_relation]:
+            grafana_relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
+
+    def remove_dashboards(self, event: RelationBrokenEvent) -> None:
+        """Remove a dashboard if the relation is broken."""
+        app_ids = _type_convert_stored(self._stored.id_mappings[event.app.name])
+
+        del self._stored.id_mappings[event.app.name]
+        for id in app_ids.keys():
```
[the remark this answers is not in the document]
simskij:
```suggestion
for id in app_ids:
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -49,6 +50,151 @@
...
+    "description": None,
+    "error": None,
+    "hide": 0,
+    "includeAll": False,
+    "label": "hosts",
+    "multi": True,
+    "name": "host",
+    "options": [],
+    "query": {
+        "query": 'label_values(up{juju_model="$juju_model",juju_model_uuid="$juju_model_uuid",juju_application="$juju_application"},host)',
```
sed-i, line 184: I wounder if we could somehow harness `class JujuTopology` for producing these repeating strings, to reduce maintenance toil.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -212,6 +358,19 @@ def _decode_dashboard_content(encoded_content: str) -> str:
     return lzma.decompress(base64.b64decode(encoded_content.encode("utf-8"))).decode()
 
 
+def _inject_dashboard_dropdowns(content: str) -> str:
+    """Make sure dropdowns are present for Juju topology."""
```
sed-i, line 362: Could you add to the docstring some more info/terms that would anchor what exactly the function does?
For example: 
- "`content` is a dashboard in a format documented in ..."
- "return an augmented dashboard with X property, if it did not exist"

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -858,6 +1019,26 @@ def remove_dashboards(self, event: RelationBrokenEvent) -> None:
         for grafana_relation in self.model.relations[self._grafana_relation]:
             grafana_relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
 
+    def _strip_existing_datasources(self, template: dict) -> dict:
+        """Remove existing reactive charm datasource templating out."""
+        for i in range(len(template["dashboard"]["templating"]["list"])):
```
sed-i: Why iterate with index and not over the objects themselves? (`for itm in template["dashboard"]["templating"]["list"]`)

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -858,6 +1019,26 @@ def remove_dashboards(self, event: RelationBrokenEvent) -> None:
         for grafana_relation in self.model.relations[self._grafana_relation]:
             grafana_relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
 
+    def _strip_existing_datasources(self, template: dict) -> dict:
+        """Remove existing reactive charm datasource templating out."""
```
sed-i, line 1024:
```suggestion
        """Remove existing reactive charm datasource templating."""
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -858,6 +1019,26 @@ def remove_dashboards(self, event: RelationBrokenEvent) -> None:
...
+        for i in range(len(template["dashboard"]["templating"]["list"])):
+            if "Juju" in template["dashboard"]["templating"]["list"][i]["datasource"]:
+                template["dashboard"]["templating"]["list"][i]["datasource"] = r"${prometheusds}"
+            if template["dashboard"]["templating"]["list"][i]["name"] == "host":
+                template["dashboard"]["templating"]["list"][i] = REACTIVE_CONVERTER
+
+        if "__inputs" in template["dashboard"]:
```
sed-i: Add comment what does it mean if there's an "__inputs"?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -858,6 +1019,26 @@ def remove_dashboards(self, event: RelationBrokenEvent) -> None:
         for grafana_relation in self.model.relations[self._grafana_relation]:
             grafana_relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
 
+    def _strip_existing_datasources(self, template: dict) -> dict:
+        """Remove existing reactive charm datasource templating out."""
+        for i in range(len(template["dashboard"]["templating"]["list"])):
+            if "Juju" in template["dashboard"]["templating"]["list"][i]["datasource"]:
+                template["dashboard"]["templating"]["list"][i]["datasource"] = r"${prometheusds}"
+            if template["dashboard"]["templating"]["list"][i]["name"] == "host":
```
sed-i: Add comment what does it mean if there's a "host"?

sed-i (dismissed):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -383,7 +384,7 @@ def add_dashboard(self, content: str) -> None:
 
         # Use as id the first chars of the encoded dashboard, so that its
         # it is predictable across units.
-        id = f"prog:{encoded_dashboard[0:7]}"
+        id = f"prog:{encoded_dashboard[-24:-16]}"
```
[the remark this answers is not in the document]
rbarry82: Simply that it's far enough away to avoid both the header and footer, which may not be unique.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: Note also that the dashboards are not using the JujuTopology class. Mostly because it did not exist when this was added, but also because `charm.name` does not have the level of uniqueness that `unit.name` does. 

Note also that the Grafana charm library/libraries are not using this anywhere other than a secondary salt for dict values. The dashboards themselves are populated from the template dropdowns, which resolve data from Prometheus labeling, and from there, Juju topology.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: That's a bridge to cross if/when we come to a point that `class JujuTopology` changes and these values are no longer sufficient.

It would really be a lot of string formatting which wouldn't gain anything. These repeating strings for e.g. 

`label_values(up{juju_model="$juju_model",juju_model_uuid="$juju_model_uuid",juju_application="$juju_application"},host)`

Basically say:

"Get the Grafana dashboard variable values for `$juju_model`, `$juju_model_uuid`, etc (set it dropdowns seen in the demo), then make a trivial Prometheus `up{...` query. Get the results, and pull the `host` label value back out, then assign it to `"name": "host"`, which will present as a `$host` variable in Grafana dashboards.

The intersection here is essentially that we are dynamic values from Prometheus back to Grafana without interacting at all. We have no need whatsoever to actually *get* the value of `juju_model` from `JujuTopology`, and doing so would be harmful, since we want it to be set from the dropdown.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: We want an iterator to manipulate the list values in place. `for item in ...` doesn't give us a good way to reinsert it without going back through again to find the index, and list values are not guaranteed to be unique anyway.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: Not to put too fine a point on it, but it means "we replace that value".

There is a point, particularly in the proxy charms, where we are operating on reverse engineering machine charms and whatever their authors may have done by seeing what comes in. The code here is, to me, self-documenting enough. "If there is a key named 'host' and its value is 'name', replace the entire node with `REACTIVE_CONVERTER` as a constant".

`host` is purely the name of a variable set by some of the reactive charms. It has no intrinsic meaning to Grafana at all, and it's not guaranteed in any way that all of the Reactive charms will even follow this pattern.

I'm not against comments, but these seem self-documenting enough, and we don't actually _know_ the reasons why these values were set.

Well, we sort-of do, kinda. That having `__inputs` as a field tells Grafana that it is a template, and it should substitute those values, but without including a substantial portion of one of the dashboards we are converting as an example (an example of which is already in the tests for the proxy itself), it won't be meaningful, since Grafana doesn't have significant documentation around automating this which we can point to anyway. Even then, not _all_ Grafana templates include an `__inputs` section. In order to include one, you must build the dashboard from the UI, export it from the UI, be in a version prior to 5.3.4, or if after 5.3.4, do not check the "external use" box. It's an edge case, just one that needs to be solved.

It's not considered logic, it is merely whack-a-mole to fit a square peg (reactive dashboards) into a round hole (LMA dashboard interfaces). If we end up with any more cases, we'll have to build what is probably an equally awful abstraction around accessing arbitrarily nested values without using `eval` through creative use of `__getitem__` or `__getattr__`, which is a path I'd rather avoid for now

In this case, "do you understand what the code does?" is as valuable as "do you understand why it was written this way"? Because the answer to why it was written this way is "because it had to be in order to convert the dashboards".

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: This comment @rbarry82 is very informative and worth copying over into the lib itself

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: Surely some part of these 3 paragraphs of clarifications could be boiled down to something informative to go into the code..?

rbarry82:
@mmanciop @simskij @dstathis @Abuelodelanada pay attention to me! This is the last part of the proxy (actually all of the changes are in that charm and none of it runs on the Grafana side, but to keep in sync so we can publish it)

sed-i (dismissed):
I unresolved a few of my comments I thought deserve a response.

rbarry82:
They were already responded to with updated docstrings.

mmanciop (dismissed):
Got this error on upgrading `grafana` with a locally-built charm when updating over the `grafana` juju application in the bundle:

```
unit-grafana-0: 18:02:46 ERROR unit.grafana/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 566, in <module>
    main(GrafanaCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 426, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 276, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 736, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 155, in _on_upgrade_charm
    self._configure()
  File "./src/charm.py", line 184, in _configure
    grafana_datasources = self._generate_datasource_config()
  File "./src/charm.py", line 525, in _generate_datasource_config
    "name": source_info["source_name"],
KeyError: 'source_name'
```

So, PR not ready to merge :-(

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+        # it's on the bucket, but using event explicitly keeps the mental model in
+        # place for reactive
+        for k in event.relation.data[event.unit].keys():
+            if k.startswith("request_"):
+                templates.append(json.loads(event.relation.data[event.unit][k])["dashboard"])
+
+        # Try app data if we haven't found it
+        for k in event.relation.data[event.app].keys():
+            if k.startswith("request_"):
+                templates.append(json.loads(event.relation.data[event.unit][k])["dashboard"])
```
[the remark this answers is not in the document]
rbarry82: This should not be `app` from a reactive charm. The dashboards are sent from units.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+    In its most streamlined usage, :class:`GrafanaDashboardAggregator` is
+    integrated in a charmed operator as follows:
+
+        self.grafana = GrafanaDashboardAggregator(self)
+
+    Args:
+        charm: a :class:`CharmBase` object which manages this
+            :class:`GrafanaProvider` object. Generally this is
+            `self` in the instantiating class.
+        target_relation: a :string: name of a relation managed by this
```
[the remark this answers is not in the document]
rbarry82: Updated

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            dashboard_templates={},
+            id_mappings={},
+        )
+
+        self._charm = charm
+        self._target_relation = target_relation
+        self._grafana_relation = grafana_relation
+
+        self.framework.observe(
+            self._charm.on[self._target_relation].relation_changed,
```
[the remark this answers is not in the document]
rbarry82: This is actually already [available in this library](https://github.com/canonical/grafana-k8s-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L147), but not used since we cannot guarantee the "direction" of legacy/reactive charms, and maximum compatibility is the goal.

If the relation is misspelled, instantiating the class will blow up from Juju. It's not a great user experience, but we're building a bridge to the past...

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            self._charm.on[self._target_relation].relation_broken,
+            self.remove_dashboards,
+        )
+
+    def update_dashboards(self, event: RelationEvent) -> None:
+        """If we get a dashboard from a reactive charm, parse it out and update."""
+        if self._charm.unit.is_leader():
+            self._upset_dashboards_on_event(event)
+
+    def _upset_dashboards_on_event(self, event: RelationEvent) -> None:
```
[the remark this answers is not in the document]
rbarry82: This is actually meant to be "upset", to match [usage elsewhere in the library](https://github.com/canonical/grafana-k8s-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L466)

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            return
+
+        self._stored.id_mappings[event.app.name] = dashboards
+
+        for id in dashboards.keys():
+            self._stored.dashboard_templates[id] = self._content_to_dashboard_object(
+                dashboards[id], event
+            )
+
+        # It's still ridiculous to add a UUID here, but needed
```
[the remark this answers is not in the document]
rbarry82: Also explained [elsewhere in the library](https://github.com/canonical/grafana-k8s-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L468), and anyone who was reading the code rather than just a diff would come across it long before this.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -753,3 +753,186 @@ def dashboards(self) -> List[Dict]:
...
+            content = _encode_dashboard_content(
+                Template(dash).render(host=event.unit.name, datasource="prometheus")
+            )
+            id = "prog:{}".format(content[-24:-16])
+
+            dashboards[id] = content
+        return {**builtins, **dashboards}
+
+    def _maybe_builtin_dashboards(self, event: RelationEvent) -> Dict:
+        """Scans the built-in dashboards and tries to match one with the event."""
```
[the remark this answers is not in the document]
rbarry82: The function name is unlikely to change, it's an artifact of Python's return typing system, but there's not a consistent wording around [optional types](https://en.wikipedia.org/wiki/Option_type) returned from functions, and `_optionally_get_builtin_dashboard` doesn't have the same feel. I updated the function name and docstring so maybe it's clearer.

sed-i (approved):

simskij (approved):

PR #46, opened 2021-12-04
Integration Tests

rbarry82:
Get integration tests up while I keep going on the rest of the charms.

This gives a first pass at review, since most of the wrappers around the Grafana API (running the action to get the password, setting headers for authenticated auth) will need to go into the "other" bundle tests anyway.

On the upside, integration tests actually discovered a bug in `GrafanaSourceProvider` which was probably hidden because the library is out of date in the Prometheus charm. When moving to the "simple" UX instantiation, `type` was no longer being passed, which would be problematic once we integrate with Loki/Agent


tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    get_grafana_dashboards,
+    get_grafana_datasources,
+    oci_image,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy_with_alternative_images(ops_test, grafana_charm):
```
sed-i, line 21:
```suggestion
async def test_build_and_deploy(ops_test, grafana_charm):
```

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+    grafana_app_name = "grafana"
+    tester_app_name = "grafana-tester"
+
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=grafana_app_name
+    )
+    await ops_test.model.deploy(
+        grafana_tester_charm, resources=tester_resources, application_name=tester_app_name
+    )
```
sed-i, line 79: In pytest test are run sequentially. Why not omit the 
```
await ops_test.model.applications[app_name].remove()
await ops_test.model.reset()
```
above so you won't need to redeploy here?
I think instead of `remove, .reset` duo we can create another `test_*.py` file, i.e. these are not unit tests, these are integration tests.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    await ops_test.model.wait_for_idle(apps=[grafana_app_name], status="active")
+    await ops_test.model.wait_for_idle(apps=[tester_app_name], status="active")
+    await ops_test.model.block_until(
+        lambda: len(ops_test.model.applications[grafana_app_name].units) > 0
+    )
+    await ops_test.model.block_until(
+        lambda: len(ops_test.model.applications[tester_app_name].units) > 0
+    )
```
sed-i, line 87: Is this a workaround for [python-libjuju/575](https://github.com/juju/python-libjuju/pull/575)?

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+    grafana_app_name = "grafana"
+    tester_app_name = "grafana-tester"
+
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=grafana_app_name
+    )
+    await ops_test.model.deploy(
+        grafana_tester_charm, resources=tester_resources, application_name=tester_app_name
+    )
```
sed-i, line 135: Same comment about resetting and redeploying: imo in itests the organization should be different than for unit tests.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    await ops_test.model.deploy(
+        grafana_tester_charm, resources=resources, application_name=app_name
+    )
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active")
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
+
+    assert ops_test.model.applications[app_name].units[0].workload_status == "active"
+
+    await ops_test.model.applications[app_name].remove()
+    await ops_test.model.reset()
```
simskij, line 57: implicitly covered by the next test

tests/integration/test_charm.py
simskij: Still don't want state between tests though. There also shouldn't be an explicit order to how the test needs to be run.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+    }
+    grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+    grafana_app_name = "grafana"
+    tester_app_name = "grafana-tester"
+
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=grafana_app_name
+    )
+    await ops_test.model.deploy(
```
simskij, line 77: gather these and give all of us some time back 😅

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+        grafana_tester_charm, resources=tester_resources, application_name=tester_app_name
+    )
+    await ops_test.model.wait_for_idle(apps=[grafana_app_name], status="active")
+    await ops_test.model.wait_for_idle(apps=[tester_app_name], status="active")
```
simskij, line 81: gather

tests/integration/test_charm.py
simskij: gather these as well

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    }
+    grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+    grafana_app_name = "grafana"
+    tester_app_name = "grafana-tester"
+
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=grafana_app_name
+    )
+    await ops_test.model.deploy(
+        grafana_tester_charm, resources=tester_resources, application_name=tester_app_name
```
simskij, line 134: gather

tests/integration/test_charm.py
simskij: There are countless sources for this a quick google away, but integration tests, just as unit tests, should be atomic as well as order-independent.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+    await ops_test.model.wait_for_idle(apps=[grafana_app_name], status="active")
+    await ops_test.model.wait_for_idle(apps=[tester_app_name], status="active")
+    await ops_test.model.block_until(
+        lambda: len(ops_test.model.applications[grafana_app_name].units) > 0
+    )
+    await ops_test.model.block_until(
+        lambda: len(ops_test.model.applications[tester_app_name].units) > 0
+    )
```
simskij, line 143: couple of gathers here as well.

simskij (dismissed):
🎉

tests/integration/test_charm.py
sed-i: With pytest, tests are run in order. For this reason I think "atomic itest" = separate file.

tests/integration/test_charm.py
sed-i: With pytest, tests are run in order. For this reason I think "atomic itest" = separate file.

tox.ini
```diff
@@ -78,7 +78,7 @@ deps =
     types-setuptools
     types-toml
 commands =
-    mypy {[vars]all_path} {posargs}
+    mypy {[vars]all_path} {posargs} --exclude {[vars]tst_path}/integration
```
sed-i, line 81: The exclude can go into `[tool.mypy]` in pyproject.

tests/integration/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+    """Ensure that the tester charm uses the current Grafana libraries."""
+    libs = [
+        Path("lib/charms/grafana_k8s/v0/", lib)
+        for lib in ["grafana_dashboard.py", "grafana_source.py"]
+    ]
+    for lib in libs:
+        Path("tests/integration/grafana-tester", lib.parent).mkdir(parents=True, exist_ok=True)
+        shutil.copyfile(
+            lib.as_posix(), "tests/integration/grafana-tester/{}".format(lib.as_posix())
+        )
```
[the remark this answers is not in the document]
rbarry82: This is essentially the same safety I raised on the Loki charm elsewhere. If this _ever_ ends up being tested with anything Windows, we get safety.

`shutil.copytree` is definitely simpler, but "fairly" safe for an additional line isn't a tradeoff I like.

.gitignore
```diff
@@ -5,3 +5,4 @@ __pycache__
 .venv
 .coverage
 venv
+tests/integration/grafana-tester/lib/charms/grafana_k8s/*
```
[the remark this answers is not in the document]
rbarry82: It also exists to provide a dummy grafana datasource.

Augmenting other operators works, just avoided for the same testing pyramid reasons. The integration tests "for this repository" don't depend on anything else by default, which provides a very trivial "does this charm build, does the most current library for a given relation _really_ work, or did the unit tests miss some bug?"

Integrating with other charms which use other API/lib versions, constructor args, or whatever is amazing, but belongs in a larger bundle test from my POV

tests/integration/grafana-tester/metadata.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Canonical Ltd.
```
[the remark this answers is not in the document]
rbarry82: This was automatically added when I did a `git add tests/integration`. Removed now.

tests/integration/grafana-tester/metadata.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: |
+  This charm is nothing more than a dummy which can present itself to Grafana
+  to pass over dashboards and datasources which do nothing.
+summary: |
+  A charm to test the Grafana operator
+platforms:
+  - kubernetes
+containers:
+  grafana-tester:
+    resource: grafana-tester-image
```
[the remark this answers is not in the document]
rbarry82: [There is for libjuju](https://github.com/juju/python-libjuju/blob/dd382ec32b56d84d95f6edb249869ee69f049157/juju/model.py#L1574). I tested removing it and it's still needed.

This could be variable in `test_charm.py` and the `oci_image` helper dropped. I don't have a strong opinion either way. But this is here because `libjuju` (via `OpsTest`) requires it, not because the Charm Metadata does. It's basically just there (and used from `test_charm -> helpers.oci_image()`) to get the parameter passed to `libjuju` without hardcoding yet another constant.

tests/integration/helpers.py
```diff
@@ -0,0 +1,205 @@
...
+        unit_num: integer number of a juju unit
+
+    Returns:
+        unit address as a string
+    """
+    status = await ops_test.model.get_status()
+    return status["applications"][app_name]["units"]["{}/{}".format(app_name, unit_num)]["address"]
+
+
+async def check_grafana_is_ready(ops_test: OpsTest, app_name: str, unit_num: int) -> bool:
```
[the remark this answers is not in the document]
rbarry82: Done.

tests/integration/helpers.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    host = await unit_address(ops_test, app_name, unit_num)
+    pw = await unit_password(ops_test, app_name, unit_num)
+    grafana = Grafana(host=host, pw=pw)
+    is_ready = await grafana.is_ready()
+    assert is_ready
+    return is_ready
+
+
+async def get_grafana_settings(ops_test: OpsTest, app_name: str, unit_num: int) -> dict:
```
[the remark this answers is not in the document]
rbarry82: The new testing framework will transparently namespace these with a folder-level module. 

There are 3 of them, ultimately, for `grafana_[datasources|dashboards|settings]`. It's not _that_ many, and if they ever get used elsewhere, even a bare import of the function will make it immediately obvious "whose" settings we're getting.

I'm not completely against changing the names, but they were named this way in pseudo-preparation for tying it into something else, and it matches the existing patterns in Prometheus.

tests/integration/test_charm.py
rbarry82: Done

tests/integration/test_charm.py
rbarry82: I think this needs discussion in the document.

On the one hand, I agree with "atomic" tests being in separate files. On the other, I'd love to avoid having 3+ `test_foo` files which have 20 lines of identical boilerplate importing helper functions and setting a license header. Certainly there is a point where splitting the tests into individual files makes sense, but for me, it's much more about "feel".

Things which are logically connected should be together. For example, a "bare" test which asks "can I build, deploy, run actions, and remove it safely without relations", and another for "can I add a dashboard relation, does the dashboard show up, does data render, etc, etc". As the test suites evolve and things like webdrivers are added, they'll probably naturally segment.

I don't think that "the behavior of the framework is sequential (for now), ordering will always be there" is a safe assumption. I also like removing not only because it keeps the tests separated/idempotent, but because we have seen plenty of "real world" bugs where there are suddenly exceptions from relations when one is broken. Leaving it deployed would leave those cases untested.

sed-i:
`tox -e static` fails on the latest revision. Can you rerun with `--recreate`?
Specifically:
```
lib/charms/grafana_k8s/v0/grafana_dashboard.py:646: error: unused "type: ignore" comment
            from jinja2 import Template  # type: ignore
            ^
lib/charms/grafana_k8s/v0/grafana_dashboard.py:647: error: unused "type: ignore" comment
            from jinja2.exceptions import TemplateSyntaxError  # type: ignore
            ^
```

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -245,11 +245,11 @@ def __init__(
         self._source_port = source_port
 
         self.framework.observe(events.relation_joined, self._set_sources)
-        self.framework.observe(refresh_event, self._set_unit_ip)
+        self.framework.observe(refresh_event, self._set_unit_ip)  # type: ignore[arg-type]
```
sed-i: I suspect that this type annotation is wrong:
`refresh_event: CharmEvents`
It should probably be `HookEvent` or `EventBase`.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -643,8 +643,8 @@ def _render_dashboards_and_emit_event(self, relation: Relation) -> None:
 
         # Import only if a charmed operator uses the consumer, we don't impose these
         # dependencies on the client
-        from jinja2 import Template
-        from jinja2.exceptions import TemplateSyntaxError
+        from jinja2 import Template  # type: ignore
+        from jinja2.exceptions import TemplateSyntaxError  # type: ignore
```
sed-i: Any idea why this addition passes on ghwf but fails locally (fresh dev dir)?

sed-i (commented):
Seems odd that you need so many `# type: ignore`s.
BTW, for some reason the changes from the latest commits do not show up in the "Files changed" tab.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: I wish I knew. This is an enormous part of the reason why I don't like `mypy`. Even with `too` and containerized environments, it's almost always different. Virtually every patch I push has something which passed locally and GitHub workflows don't like, so I have to add `#type: ignore` comments to pass GHWF which make it a 100% reproducible failure locally.

These are actually already in the repo, and the only reason they showed up at all is that I forgot that this happens in GHWF and made linting pass locally.

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: This leads to problems the other way, for anyone who is using a language server and tries to throw in a parent class, they will immediately have their IDE complain _if_ it's any kind of CharmEvent. It's six of one and half a dozen of the other, but at least keeping at `CharmEvent` hopefully gives an indication to library users about what kind of events we _think_ should be used.

sed-i: I have a different understanding: CharmEvents is an aggregator, not an ancestor.
And relation events are not part of CharmEvents.

rbarry82: Just so we're on the same page, [RelationEvents from charms are CharmEvents](https://github.com/canonical/operator/blob/d60f6d6a7e0dba9f6c69c7778ba9d0d355e788c5/ops/charm.py#L635), conceptually. While `RelationChangedEvent` (for example) is not the same class as `CharmEvents`, and becomes a `BoundEvent` through an `EventSource` via [define_event](https://github.com/canonical/operator/blob/824aa2d8996ea548c913317c2df6bac258f0737b/ops/framework.py#L389), which eventually re-binds the "real" class, the point in annotating it that way is conceptual for people who may actually be using the library.

The "magic" in `define_event` rebinding/re-emitting through the framework writing/sending files confuses `mypy`, and I'd rather have clarity about what kind of things this is intended to be used for than be technically correct for `mypy`. Incidentally, this is just another place where `mypy` is inconsistent to the point of frustration. The same library does [the same thing](https://github.com/canonical/grafana-k8s-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/lib/charms/grafana_k8s/v0/grafana_source.py#L319) elsewhere, which `mypy` doesn't care about.

I'd opt for UX and clarify over making a schizophrenic linter happy.

tests/integration/conftest.py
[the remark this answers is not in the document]
sed-i: I'm probably missing something but `shutil.copytree` takes posix paths. Could you point me to the potential issue with Windows?
Feel free to resolve this.

.gitignore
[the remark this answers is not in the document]
sed-i: Tester charms (and mocks in general) introduce maintenance overhead and divergence from production behavior.
I am not convinced no-deps justifies having the tester charm but I will not block this PR over it.

metadata.yaml
```diff
@@ -39,3 +39,4 @@ resources:
   grafana-image:
     type: oci-image
     description: upstream docker image for Grafana
+    upstream-source: docker.io/ubuntu/grafana@sha256:8952155a72868b609e6b2e5ac5d20c540b232380f17dcf8aa60ac04b234edf9c
```
sed-i, line 42: I recall having decided to use human-readable multi-arch tags rather than hashes..?

pyproject.toml
```diff
@@ -28,7 +28,7 @@ profile = "black"
 max-line-length = 99
 max-doc-length = 99
 max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
+exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv", "tests/integration"]
```
sed-i, line 31: I wonder why excluding itests from linting was needed

tox.ini
```diff
@@ -89,5 +89,16 @@ deps =
...
+[testenv:integration]
+description = Run integration tests
+deps =
+    aiohttp
+    asyncstdlib
+    git+https://github.com/juju/python-libjuju.git
+    pytest
+    git+https://github.com/charmed-kubernetes/pytest-operator.git
+commands =
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {toxinidir}/tests/integration
```
sed-i, line 104:
```suggestion
    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tst_path}/integration
```

sed-i (approved):

pyproject.toml
rbarry82: Primarily, it hates the redefinition of `charm.py`, even though it should be under a different module-level namespace.

metadata.yaml
rbarry82: I recall the opposite, so we could uniquely identify _exactly_ which commit broke it. Let's discuss tomorrow at the meeting so we're all on the same page. I'm happy to change it.

.gitignore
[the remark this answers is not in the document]
rbarry82: Yes, but it's the same testing strategy pyramid.

The per-repo integration tests are not intended to mimic production. That's what bundle/e2e tests are for. They're meant to demonstrate that, not only do the unit tests demonstrate that it works in code, but that we actually get the correct result on an actual Juju run.

For the purposes of the single charm integration test, this is hopefully sufficient. Like the other comment, I'm happy to discuss at the meeting, but I recall this being the agreement.

tests/integration/conftest.py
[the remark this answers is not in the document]
rbarry82: It's essentially the same reasoning as https://github.com/canonical/loki-k8s-operator/issues/47#issue-1071035515, and using `pathlib` everywhere keeps it consistent.

For one, `shutil` doesn't actually guarantee POSIX paths anywhere, so it surely could be Windows, but the concern is pretty much in edge cases. `pathlib.Path` keeps us honest instead of trying to mangle filenames at any point. The concern is not just Windows, but also if these go into a "bundle" as-is. If we use `copytree` and there are relative paths pointing to a broken link somewhere (`../../foo`), we could inadvertently copy something unintended or a massive directory tree.

That's not a big security concern, really, but it could definitely cause a headache for anyone trying to run the tests locally, or cause unrelated test failures with obscure messages. Being explicit about what we're copying is just putting on guardrails

rbarry82 (commented):
Updated inline

.gitignore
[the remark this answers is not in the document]
simskij: > The per-repo integration tests are not intended to mimic production. That's what bundle/e2e tests are for. They're meant to demonstrate that, not only do the unit tests demonstrate that it works in code, but that we actually get the correct result on an actual Juju run.

Fully agree.

sed-i (approved):

PR #48, opened 2021-12-16
Ensure that old source keys with hyphens are remapped on upgrade

rbarry82:
Make sure the the replacement of `-` with `_` from https://github.com/canonical/grafana-k8s-operator/pull/45 actually reflects in keys on upgrade


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -436,6 +436,19 @@ def _remove_source(self, source_name: str) -> None:
         """Remove a datasource by name."""
         self._stored.sources_to_delete.add(source_name)
 
+    def upgrade_keys(self) -> None:
+        """On upgrade, ensure stored data maintains compatibility."""
```
sed-i, line 440: Reminds me of [`TODO_BEFORE`](https://www.fluentcpp.com/2019/01/01/todo_before-clean-codebase-2019/) macro: we should have some kind of reminder to drop this logic in a future version.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -436,6 +436,19 @@ def _remove_source(self, source_name: str) -> None:
...
+        # self._stored.sources may have hyphens instead of underscores in key names.
+        # Make sure they reconcile.
+        sources = _type_convert_stored(self._stored.sources)
+        for rel_id in sources.keys():
+            for i in range(len(sources[rel_id])):
+                sources[rel_id][i].update(
+                    {k.replace("-", "_"): v for k, v in sources[rel_id].items()}
```
sed-i, line 447: Should we add an assertion in one of the integration tests that traverses relation data and makes sure everything is sanitized according to some spec?

sed-i (dismissed):

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: Aiming for API stability before changing for style (and having tests) is definitely a goal.

Realistically, I suspect that we will either need to start implementing some kind of key which includes metadata versions in the relation/peer/etc database and/or stored state in the same way as migrations work in Django/Rails.

It would be great to drop this in the future, but any cutoff date after which we can say "ok, we don't need to do this anymore because it's unsupported" is up someone else. I would expect the upgrade logic per charm to get more complicated if the API continues to change, not less.

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: Done.

tests/test_source_consumer.py
```diff
@@ -280,6 +280,38 @@ def test_consumer_noop_on_source_removal_if_not_leader(self):
...
+                {
+                    "source-name": "shouldconvert",
+                    "source_name": "shouldconvert",
+                    "source-type": "prometheus",
+                    "source_type": "prometheus",
+                    "unit": "prometheus/0",
+                    "url": "1.2.3.4",
+                }
+            ]
+        }
```
sed-i, line 305: Why keep the old keys (+duplicated values) after `upgrade_keys`?

sed-i (approved):

tests/test_source_consumer.py
simskij: I don't know the exact reason why Ryan chose to keep it, but regardless: because we need to. A user might want to downgrade to the previous version. If we remove all the old keys, that won't be an option.

Abuelodelanada (approved):

PR #50, opened 2022-01-12
Self monitoring

rbarry82:
Add alertmanager rules, and send the metrics (already enabled out of the box) back to Prometheus so the alerts can actually fire.

Grafana includes a dashboard for itself which we can use.


tox.ini
```diff
@@ -78,7 +78,7 @@ deps =
     types-setuptools
     types-toml
 commands =
-    mypy {[vars]all_path} {posargs} --exclude {[vars]tst_path}/integration
+    mypy {[vars]all_path} {posargs} --exclude {[vars]tst_path}integration
```
sed-i, line 81: I made changes to this file. Have you rebased?
https://github.com/canonical/grafana-k8s-operator/pull/49/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449

I think the `/` needs to stay because I removed it from the definition of tst_path.

sed-i (approved):
I wonder if self monitoring would have a noticeable impact on the load test / dcatasheet.

metadata.yaml
```diff
@@ -32,6 +32,9 @@ requires:
   database:
     interface: db
     limit: 1
+provides:
+  metrics-endpoint:
```
mmanciop, line 36: I am thinking that, due to the nature of our charms being themselves monitoring tools, we need to put extra care in the naming, and probably break convention. We must have a discussion in the team about how to name these. cc @simskij

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,10 @@
+alert: GrafanaUnitIsUnavailable
+expr: up{%%juju_topology%%} < 1
```
mmanciop, line 2: This is for example something that, if we go ahead like @sed-i pushes for and accept to drop the support for `%%juju_topology%%`, also breaks :-)

dstathis (approved):

metadata.yaml
simskij: > I am thinking that, due to the nature of our charms being themselves monitoring tools, we need to put extra care in the naming, and probably break convention. We must have a discussion in the team about how to name these. cc @simskij

Got it! Added it to the team meeting agenda for next week.

src/prometheus_alert_rules/unit_unavailable.rule
simskij: We are not dropping that support, so this is not an issue anymore.

metadata.yaml
sed-i:
```yaml
provides:
  custodiet-ipsos-custodes::
```

dstathis (dismissed):

src/prometheus_alert_rules/unit_unavailable.rule
mmanciop, line 2: We do not need the `%%juju_topology%%` anymore, thanks to `promql_transform`

mmanciop (changes requested):

mmanciop:
Also, can I please get a matching PR in https://github.com/canonical/cos-lite-bundle ? :-)

simskij:
[9 votes for `self-*`](https://discourse.charmhub.io/t/how-should-we-call-relations-for-monitoring-cos-charms/5950), let's go with that and only use it for the ones that also provide that relation organically, ie. `self-metrics-endpoint` for Prometheus, `self-logs` for Loki, `self-logs` for Grafana Agent, and `self-dashboards` for Grafana itself.

rbarry82:
And so it was written

src/prometheus_alert_rules/unit_unavailable.rule
simskij: See https://github.com/canonical/cos-lite-bundle/issues/24

src/charm.py
```diff
@@ -95,6 +96,16 @@ def __init__(self, *args):
         self._grafana_datasources_hash = None
         self._stored.set_default(k8s_service_patched=False, admin_password="")
 
+        # -- Prometheus self-monitoring
```
sed-i, line 99:
```suggestion
        # -- Self-monitoring
```

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,10 @@
...
+for: 0m
+labels:
+  severity: critical
```
sed-i, line 5: Do we need a `self-monitoring` label here?

src/prometheus_alert_rules/unit_unavailable.rule
sed-i, line 2: Do we need some sort of a low-pass filter here? e.g.
```suggestion
expr: avg_over_time(up{%%juju_topology%%}[1m]) < 0.5
```
or perhaps something using [`changes`](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes)?

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: Why would we?

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: Why would we average this?

simskij: if it's down, it's down. but we should probably have a somewhat longer `for` than 0m to avoid false positives.

src/prometheus_alert_rules/unit_unavailable.rule
simskij: No. That would just increase the cardinality. And remember: the rule is not only for bundle deployments, so it's not necessarily always self-monitoring.

simskij (dismissed):

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: The question then becomes: "what do we think is an acceptable middle ground?"

If there's more than one Grafana unit (which would be a different rule), do we want to alert if one is down for 5 minutes? 10 minutes?

If there's only one, how long do we wait? Prometheus polling is every 15 seconds (plus the default HTTP timeout before the scrape actually fails). On the one hand, this isn't very long. On the other hand, a "busy" Grafana instance can result in an awful lot of user reports if it's down for multiple minutes, and sending an alert _before_ tickets start rolling in is always the goal.

What do we think is a good mix?

sed-i: Maybe this is a case for the severity label and multiple alerts in the group?
- one alert with low severity for `up{%%juju_topology%%} < 1` - this one is also something that MLOps would be happy to consume
- another alert with critical severity for `avg_over_time(up{%%juju_topology%%}[1m]) < 0.5`
Tickets will continue to roll until the issue is resolved, not until it is first seen, on which alert fatigue would have an impact too.

sed-i (approved):

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: Done.

balbirthomas:
Looks like integration tests got hung for 6 hours. This looks like the hang on installing agent Juju flayness.

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,22 @@
+groups:
+  alert: GrafanaUnitIsUnavailable
+  expr: up{} < 1
+  for: 0m
+  labels:
+    severity: critical
```
sed-i, line 6: Should this be one level lower than `GrafanaUnitIsDown`?

```suggestion
    severity: error
```

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,22 @@
+groups:
+  alert: GrafanaUnitIsUnavailable
+  expr: up{} < 1
```
sed-i, line 3: Just curious why the empty brackets.

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,22 @@
+groups:
```
sed-i, line 1: This does not seem like a [valid rule file](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/) - a hierarchy level is missing:
```yaml
- name: example
  rules:
```
Add itest? Not sure what the best way to make grafana unavailable though. Firewall rule?

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,22 @@
...
+    severity: critical
+  annotations:
+    summary: Grafana unit {{ $labels.juju_model }}/{{ $labels.juju_unit }} unavailable
+    description: >
+      The Grafana unit {{ $labels.juju_model }} {{ $labels.juju_unit }} is unavailable
+      LABELS = {{ $labels }}
+
+  alert: GrafanaUnitIsDown
+  expr: avg_over_time(up{}[1m]) < 0.5
+  for: 0m
```
sed-i, line 15: I suspect this combo would cause the alert to fire right away if the very first scrape fails.
The easiest (but not most accurate) solution would be 
```suggestion
  expr: avg_over_time(up{}[1m]) < 0.5
  for: 1m
```

tox.ini
```diff
@@ -94,8 +94,8 @@ description = Run integration tests
 deps =
     aiohttp
     asyncstdlib
-    juju
+    git+https://github.com/juju/python-libjuju.git@9ccadfa431b9224ce2cf4b8cc2fd26935060a379
     pytest
-    git+https://github.com/charmed-kubernetes/pytest-operator.git
+    git+https://github.com/charmed-kubernetes/pytest-operator.git@67e6eaebf745b40065167e3b599314cb9e1c674a
```
sed-i, line 99: Should probably use version numbers, if that actually works for this charm (works for me in other charms)
```ini
    juju==2.9.7
    pytest-operator==0.15.0
```

dstathis (approved):

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: That's the goal.

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: We're automatically injecting topology into the alerts, aren't we?

src/prometheus_alert_rules/unit_unavailable.rule
rbarry82: Skipping integration tests until other pieces are present, and it can be tested as a sum. This is just the first.

I did miss the name, though...

tox.ini
rbarry82: The goal is basically to stay on HEAD unless there's a breaking change which forces a specific commit, like we have right now (until libjuju pushes again)

src/prometheus_alert_rules/unit_unavailable.rule
sed-i: Not sure what you mean. (This is a minor comment so no big deal anyway)
My impression was:
1. The alert that fires right away is `GrafanaUnitIsUnavailable`, with `severity: error`.
2. `GrafanaUnitIsDown` fires only after it up is flapping (or completely down) for a minute, with `severity: critical`.

sed-i (dismissed):

simskij (approved):

simskij (approved):

sed-i (commented):
Why commit promql-transform? I though we said we'd rather depend on internet than commit binaries.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1047,7 +1047,7 @@ def _on_grafana_peer_changed(self, _: RelationChangedEvent) -> None:
             return
         self.on.dashboards_changed.emit()
 
-    def update_dashboards(self, relation: Optional[Relation] = None) -> None:
+    def update_dashboards(self, relation: Relation = None) -> None:
```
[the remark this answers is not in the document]
rbarry82: It would be bad for airgapping, but this will get the same thing as Prometheus anyway, and this will be removed.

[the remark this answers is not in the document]
simskij: Agreed, but rather than pushing binaries, we should consider priming the charm with promql-transform while packing.

src/prometheus_alert_rules/unit_unavailable.rule
simskij: Brackets are not really needed for promql-transform to work, but they don't hurt either. 😅

Abuelodelanada (approved):

metadata.yaml
simskij: It is still correct. `self-` we decided to only add to the provider-side of something that also is a requirer of the same interface. For instance, for Prometheus `metrics-endpoint` is the requirer, while `self-metrics-endpoint` is the provider.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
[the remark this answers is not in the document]
simskij: @rbarry82 can we please resolve this prior to merging? If we merge with the binary in there, it will be stuck in the reflog unless purged manually, even if we delete it in a subsequent commit, which is something I'd rather not have us do.

balbirthomas (approved):

PR #53, opened 2022-01-21
Purge relation data from `GrafanaDashboardProvider` if the dashboard path cannot be resolved or is empty

rbarry82:
To enable `cos-config` and potentially other users, add an arg
for an optional extra event to listen to to refresh dashboards

cf. https://github.com/canonical/prometheus-k8s-operator/pull/208, https://github.com/canonical/loki-k8s-operator/pull/76

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -41,7 +41,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -444,7 +444,7 @@ class GrafanaDashboardProvider(Object):
         charm: CharmBase,
         relation_name: str = DEFAULT_RELATION_NAME,
         dashboards_path: str = "src/grafana_dashboards",
-        refresh_event: Optional[HookEvent] = None,
+        refresh_events: Optional[List[HookEvent]] = None,
     ) -> None:
         """API to provide Grafana dashboard to a Grafana charmed operator.
 
@@ -495,8 +495,8 @@ class GrafanaDashboardProvider(Object):
                 where dashboard templates can be located. By default, the library
                 expects dashboard files to be in the `<charm-py-directory>/grafana_dashboards`
                 directory.
-            refresh_event: a :class:`HookEvent` event on which the dashboard path should
-                address should be refreshed.
+            refresh_events: an optional list of :class:`HookEvent` event(s) on which the
+                dashboard path should be refreshed.
         """
         _validate_relation_by_interface_and_direction(
             charm, relation_name, RELATION_INTERFACE_NAME, RelationRole.provides
@@ -521,8 +521,9 @@ class GrafanaDashboardProvider(Object):
         self.framework.observe(self._charm.on.leader_elected, self._update_all_dashboards_from_dir)
         self.framework.observe(self._charm.on.upgrade_charm, self._update_all_dashboards_from_dir)
 
-        if refresh_event:
-            self.framework.observe(refresh_event, self._update_all_dashboards_from_dir)  # type: ignore[arg-type]
+        refresh_events = refresh_events or []
+        for e in refresh_events:
+            self.framework.observe(e, self._update_all_dashboards_from_dir)  # type: ignore[arg-type]
 
         self.framework.observe(
             self._charm.on[self._relation_name].relation_created,
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -577,21 +577,24 @@ class GrafanaDashboardProvider(Object):
             for dashboard_relation in self._charm.model.relations[self._relation_name]:
                 self._upset_dashboards_on_relation(dashboard_relation)
 
-    def _update_all_dashboards_from_dir(self, _: HookEvent) -> None:
+    def _update_all_dashboards_from_dir(
+        self, _: Optional[HookEvent] = None, dir: Optional[str] = ""
+    ) -> None:
         """Scans the built-in dashboards and updates relations with changes."""
         # Update of storage must be done irrespective of leadership, so
         # that the stored state is there when this unit becomes leader.
 
         # Ensure we do not leave outdated dashboards by removing from stored all
         # the encoded dashboards that start with "file/".
-        if self._dashboards_path:
+        dashboards_path = dir or self._dashboards_path
+        if dashboards_path:
             stored_dashboard_templates = self._stored.dashboard_templates
 
             for dashboard_id in list(stored_dashboard_templates.keys()):
                 if dashboard_id.startswith("file:"):
                     del stored_dashboard_templates[dashboard_id]
 
-            for path in filter(Path.is_file, Path(self._dashboards_path).glob("*.tmpl")):
+            for path in filter(Path.is_file, Path(dashboards_path).glob("*.tmpl")):
                 id = "file:{}".format(path.stem)
                 stored_dashboard_templates[id] = self._content_to_dashboard_object(
                     _encode_dashboard_content(path.read_bytes())
@@ -600,6 +603,24 @@ class GrafanaDashboardProvider(Object):
             if self._charm.unit.is_leader():
                 for dashboard_relation in self._charm.model.relations[self._relation_name]:
                     self._upset_dashboards_on_relation(dashboard_relation)
+
+    def reload_dashboards_from_dir(self, dir: Optional[Union[Path, str]] = None) -> None:
+        """Triggers a reload of dashboard outside of an eventing workflow.
+
+        Args:
+            dir: an optional `Path` or `str` specifying the directory to check.
+        """
+        dir = str(dir) if dir else self._dashboards_path
+        try:
+            _resolve_dir_against_charm_path(self._charm, dir)
+            self._update_all_dashboards_from_dir(dir=dir)
+
+        except InvalidDirectoryPathError as e:
+            logger.warning(
+                "Invalid Grafana dashboards folder at %s: %s",
+                e.grafana_dashboards_absolute_path,
+                e.message,
+            )
 
     def _on_grafana_dashboard_relation_created(self, event: RelationCreatedEvent) -> None:
         """Watch for a relation being created and automatically send dashboards.
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -444,6 +444,7 @@ class GrafanaDashboardProvider(Object):
         charm: CharmBase,
         relation_name: str = DEFAULT_RELATION_NAME,
         dashboards_path: str = "src/grafana_dashboards",
+        refresh_event: Optional[HookEvent] = None,
     ) -> None:
         """API to provide Grafana dashboard to a Grafana charmed operator.
 
@@ -494,6 +495,8 @@ class GrafanaDashboardProvider(Object):
                 where dashboard templates can be located. By default, the library
                 expects dashboard files to be in the `<charm-py-directory>/grafana_dashboards`
                 directory.
+            refresh_event: a :class:`HookEvent` event on which the dashboard path should
+                address should be refreshed.
         """
         _validate_relation_by_interface_and_direction(
             charm, relation_name, RELATION_INTERFACE_NAME, RelationRole.provides
@@ -517,6 +520,9 @@ class GrafanaDashboardProvider(Object):
 
         self.framework.observe(self._charm.on.leader_elected, self._update_all_dashboards_from_dir)
         self.framework.observe(self._charm.on.upgrade_charm, self._update_all_dashboards_from_dir)
+
+        if refresh_event:
+            self.framework.observe(refresh_event, self._update_all_dashboards_from_dir)  # type: ignore[arg-type]
 
         self.framework.observe(
             self._charm.on[self._relation_name].relation_created,
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -444,7 +444,6 @@ class GrafanaDashboardProvider(Object):
         charm: CharmBase,
         relation_name: str = DEFAULT_RELATION_NAME,
         dashboards_path: str = "src/grafana_dashboards",
-        refresh_events: Optional[List[HookEvent]] = None,
     ) -> None:
         """API to provide Grafana dashboard to a Grafana charmed operator.
 
@@ -495,13 +494,12 @@ class GrafanaDashboardProvider(Object):
                 where dashboard templates can be located. By default, the library
                 expects dashboard files to be in the `<charm-py-directory>/grafana_dashboards`
                 directory.
-            refresh_events: an optional list of :class:`HookEvent` event(s) on which the
-                dashboard path should be refreshed.
         """
         _validate_relation_by_interface_and_direction(
             charm, relation_name, RELATION_INTERFACE_NAME, RelationRole.provides
         )
 
+        dashboards_path = ""
         try:
             dashboards_path = _resolve_dir_against_charm_path(charm, dashboards_path)
         except InvalidDirectoryPathError as e:
@@ -519,11 +517,7 @@ class GrafanaDashboardProvider(Object):
         self._stored.set_default(dashboard_templates={})
 
         self.framework.observe(self._charm.on.leader_elected, self._update_all_dashboards_from_dir)
-        self.framework.observe(self._charm.on.upgrade_charm, self._on_upgrade_charm)
-
-        refresh_events = refresh_events or []
-        for event in refresh_events:
-            self.framework.observe(event, self._update_all_dashboards_from_dir)  # type: ignore
+        self.framework.observe(self._charm.on.upgrade_charm, self._update_all_dashboards_from_dir)
 
         self.framework.observe(
             self._charm.on[self._relation_name].relation_created,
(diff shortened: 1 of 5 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -494,6 +495,8 @@ def __init__(
                 where dashboard templates can be located. By default, the library
                 expects dashboard files to be in the `<charm-py-directory>/grafana_dashboards`
                 directory.
+            refresh_event: a :class:`HookEvent` event on which the dashboard path should
+                address should be refreshed.
```
sed-i, line 499:
```suggestion
                be refreshed.
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -444,6 +444,7 @@ def __init__(
         charm: CharmBase,
         relation_name: str = DEFAULT_RELATION_NAME,
         dashboards_path: str = "src/grafana_dashboards",
+        refresh_event: Optional[HookEvent] = None,
```
sed-i, line 447: Should this be `Optional[Sequence[HookEvent]]`?

sed-i (approved):

simskij (approved):

sed-i:
@dstathis can we do the same thing for `LokiPushApiConsumer` as part of https://github.com/canonical/loki-k8s-operator/pull/74 ?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: The reason to have a list is to register both update status and the git-sync pebble-ready, in addition to config-changed.

rbarry82: Sure, done.

sed-i (dismissed):

mmanciop (commented):
I am sure this has been discussed before, but wouldn’t it make more sense, rather than hard-coding events, to expose a method instead? It would be more flexible  and less obscure.

rbarry82:
It would be used exactly to load "whatever", though we go through the same `_resolve_dir_against_charm_path` dance.

The user story here being, more or less, "I want to load Grafana dashboards from more than one path since my git hierarchy is complex and/or I'm optionally enabling some"

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -594,6 +604,24 @@ def _update_all_dashboards_from_dir(self, _: HookEvent) -> None:
...
 
+    def reload_dashboards_from_dir(self, dir: Optional[Union[Path, str]] = None) -> None:
+        """Triggers a reload of dashboard outside of an eventing workflow.
+
+        Args:
+            dir: an optional `Path` or `str` specifying the directory to check.
+        """
+        dir = str(dir) if dir else self._dashboards_path
+        try:
+            _resolve_dir_against_charm_path(self._charm, dir)
```
sed-i, line 615:
```suggestion
            dir = _resolve_dir_against_charm_path(self._charm, dir)
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -570,30 +601,57 @@ def update_dashboards(self) -> None:
...
+        try:
+            dir = _resolve_dir_against_charm_path(self._charm, dir)
+            self._update_all_dashboards_from_dir(dir=dir)
+
+        except InvalidDirectoryPathError as e:
+            logger.warning(
+                "Invalid Grafana dashboards folder at %s: %s",
+                e.grafana_dashboards_absolute_path,
+                e.message,
+            )
```
sed-i, line 653: In addition to the warning, relation data should be emptied out.
cf https://github.com/canonical/prometheus-k8s-operator/pull/209
Currently I have dashboard data persist even after I removed the folder (cos-config changed config).

rbarry82: This will break the workflow for Grafana dashboards entirely due to message passing of events between provider and consumer. See `on_grafana_dashboard_relation_changed` on the provider side. An event is created from the provider side of the relation which can inform the charm "hey, that dashboard template you sent across the relation was not valid". 

This ensures that providers don't need to install Jinja or other dependencies, and this information would be completely lost.

What's the expected workflow here?

* _CURRENTLY_ If it's re-initialized with a different path, all of the old dashboards are deleted as long as the new path exists at all.
* _CURRENTLY_ if it's re-initialized with an empty path, `self._dashboards_path` is not set due to the same exception in the constructor (which this patch does not change). Dashboard data is removed from the consumer side on `RelationBroken`.
* _WITH THIS PATCH_ the behavior is effectively identical. 

A small exception handler function, ala:

```python
    for dashboard_id, data in self._stored.dashboard_templates:
        if dashboard_id.startswith("file:") and data["path"] == dir:
            del self._stored.dashboard_templates[dashboard_id]
    self._upset_dashboards_on_relation(...
```

Could handle this. But, to be clear, what you're experiencing is not behavior changed by this patch. "If the dashboard directory disappears from under the charm after working before, destroy them all" is a new requirement, so let's be clear on what the user story is

sed-i: I know the behavior I'm asking for is not related to this patch.
I didn't realize the situation is so complex so I thought it could have been addressed here by a simple removal :sweat_smile: 

The user story is: cos-config admin sets up a repo with dashboards and forms a relation with grafana; then makes a change without removing the relation, for example: unset the repo url or change to a different branch in the same repo, or changes one of those params to an invalid value such that cloning fails.

Not sure we need that user story nor if it makes sense to address it in this PR. @mmanciop ?

rbarry82: This is the patch for it, IMO.

So the question basically becomes:

Should `cos-config` be responsible for handling its own behavior there? Why would we even call this method if the path is empty?

And do we actually want to destroy monitoring rules because of a "fat finger" (invalid value)? IMO, this is one of the worst things you can do to a monitoring environment. "Well, we pushed a bad config and didn't get any warning about it because, when we made this change, all of our warnings deleted themselves" is not a good experience.

I understand that, if there's a new _valid_ path, we could do this. This would more or less make the logic of this method into "add an additional path to scan for rules changes every time there's an event", with a corresponding "remove this path".

Then, we could send an event back telling the charm author "this is an invalid loki/alertmanager path" and let them make decisions based on how it actually _should_ operate for a safe monitoring environment rather than having the default behavior as deletion. 

Deletion should be intentional IMO, in database systems, source control, monitoring systems, IPAM, AAA, and on and on. It's critical infrastructure.

sed-i: I agree, but there are a few subtleties I can think of that complicate things:
1. The user clears a config value, for example: `juju config cos-conf repo_url=`. In this case it is not clear if it is a fat finger or intentional. There is no default value for the repo url so it must be user specified and as a result it can be cleared.
2. Say the user has three branches (policy-x, policy-y, policy-z) they use throughout the different phases of an incident. There is no way for that charm to know if a fat finger instructed a switch from x to z instead of from x to y. The charm just needs to act on it.
3. Even if the charm detects that the newly provided repo url or branch name does not exist, the charm cannot change its own config, so if we do not act on the new config then we find ourselves in an inconsistent state.

rbarry82: I don’t think those complicate things, really. Destructive operations should be intentional, not default library behavior. 

1. This is the same as “are you sure”? A two step operation wherein the config value is intentionally changed, then some action is invoked to purge the data is safer. 

2. Same as 1, really. If the new branch doesn’t have dashboards, the administrator should be aware of this. If they are not, sending an event back to the charm from which a status can be set indicating the action to take for “are you sure?” can be performed. 

3. The same as the two above. 

In all cases, responding with an event which says “something went wrong, administrator intervention required to confirm whether this was intended” is safe default library behavior, and it prevents other users of the library from accidentally destroying data in charms which are _not_ `cos-config`. 

Again, it’s trust. I have both been the person who has accidentally deleted production data/wiped servers when a broken link failed to resolve, and the person who has gotten (or worse, not gotten until someone reports that a production environment went down long after a disk failure didn’t report) these alerts. 

In all cases, “are you sure you meant to do this?” would have been minimal extra effort for much added peace of mind would have been worth it.

sed-i: Again, I agree with all these but there is a problem: a charm cannot interrupt a config change; afaik there is nowhere we can place an "are you sure" - the config is changed, period, and now the charm needs to live with it. Emitting an event that would change the status to "potentially unintentional change" is good, but orthogonal.

sed-i: I suppose we could keep a "last known good configuration" in stored state but that should prompt more deliberate speccing.

sed-i: I suppose we could keep a "last known good configuration" in stored state but that would call for more deliberate speccing.

mmanciop: While I appreciate the sentiment of avoiding unintentional destructive changes, that is not part of the Juju workflow. So, if the admin, misguided or otherwise, says cos-config "Fetch from this EXISTING, EMPTY branch", we MUST oblige and wipe dashboards. If the Juju admin says "fetch from this branch that does not exist", cos config MUST wipe its current dashboards (the Juju admin specified a different branch than the current) and stuff and the Juju application goes blocked mode (new branch does not exist).

As a mitigation for these concerns related with dashboards, the impact of a dashboard being wiped is limited to people not been able to look at it: it has no impact to the telemetry we collect or alerts we issue. Obviously, alert rules for Prometheus and Loki ARE impacted by a bad config passed to cos-config by the Juju admin, and that will cause alerts not to fire. In the future, when we decide to support [recording rules](https://www.prometheus.io/docs/prometheus/latest/configuration/recording_rules/), it might cause loss of telemetry (specifically: the one not created by the recording rules), but that's how it is.

rbarry82: > Again, I agree with all these but there is a problem: a charm cannot interrupt a config change; afaik there is nowhere we can place an "are you sure" - the config is changed, period, and now the charm needs to live with it (i.e. act on it or ignore and be inconsistent with the config). Emitting an event that would change the status to "potentially unintentional change" is good, but orthogonal.

You're right. Hooks cannot be interrupted, but charms aren't an interactive system as-is in that manner anyway (for that matter, neither is `rpm`, where even printing output is bad practice). "Update the status with a message telling you what should be done to unblock this" is absolutely idiomatic, though, and it's done in a large number of charms, including our own proxy, where we update the status to "hey, you haven't set up other half of this relation pair, please do so to unblock and clear this message". 

A lot of the older reactive charms _frequently_ update the status. This is patently [the use case for BlockedStatus](https://discourse.charmhub.io/t/status-values/1168): "hey admin! we're blocked until you do something to clarify the situation."

> While I appreciate the sentiment of avoiding unintentional destructive changes, that is not part of the Juju workflow. So, if the admin, misguided or otherwise, says cos-config "Fetch from this EXISTING, EMPTY branch", we MUST oblige and wipe dashboards. If the Juju admin says "fetch from this branch that does not exist", cos config MUST wipe its current dashboards (the Juju admin specified a different branch than the current) and stuff and the Juju application goes blocked mode (new branch does not exist).

This is what I mean. The `cos-config` charm goes into `BlockedStatus` and requires intervention to unblock on a config change. What I'm pushing for is exactly the same: if there is an exception thrown, we shouldn't destroy data unless `purge=True` or some other "`--yes-i-really-really-mean-it`" (to use the Juju flag terminology) is used. This is the TLDR of my comments.

> As a mitigation for these concerns related with dashboards, the impact of a dashboard being wiped is limited to people not been able to look at it: it has no impact to the telemetry we collect or alerts we issue.

My concern is really with the whole pattern, much more with alert rules than dashboards. Dashboards are an end-user experience. This is just one of the PRs, and where the discussion happened. "I set a bad config value and `cos-config` is going to go into `BlockedStatus` anyway" should not result in "wiping" alert rules.

It can wipe dashboards. That part doesn't matter, except insofar as our libraries should behave the same way, and `cos-config` should interact with them the same way, so dashboards should behave the same way as alerts. That is: if `cos-config` is going to go into `BlockedStatus` for manual intervention because something went wrong, short-circuiting early/not wiping alert rules is the closest we'll come to a transactional operation right now.

> Obviously, alert rules for Prometheus and Loki ARE impacted by a bad config passed to cos-config by the Juju admin, and that will cause alerts not to fire. In the future, when we decide to support recording rules, it might cause loss of telemetry (specifically: the one not created by the recording rules), but that's how it is.

The charm must act on it, yes, but that doesn't necessarily mean it succeeds without intervention. It tells the admin "run this action", and that action calls the functions with some parameter which forces the operation, then it sets it back to `ActiveStatus`. The same function gets called as the refresh event anyway, so the charm acts on it _then_, once it has confirmation.

We could do better about stopping users from shooting themselves in the foot. 

Where I take issue with the patch is this being behavior in libraries which are *not* `cos-config` defaulting to "self-destruct on exception" rather than it being an optarg `cos-config` can pass in (or logic `cos-config` can handle).

Putting this bad kind of deadman's switch into a bunch of our libraries to enable a single user story -- one which could _easily_ pass an optarg -- is a bad behavior.

Loki [already passes an event back](https://github.com/canonical/loki-k8s-operator/blob/f52629305796d8e341d01165f245b06f26b05e8f/lib/charms/loki_k8s/v0/loki_push_api.py#L1090), so at least authors can listen to it an display a message.

Right now, we have 3 libraries doing this 3 ways:

* Grafana [guards against the path being empty](https://github.com/canonical/grafana-k8s-operator/blob/2a3c9314dd1bfa2d8daefa5213483551fb1302af/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L580) and won't update data if we got `InvalidDirectoryPathData`
* Loki emits an event which can be listened to, then continues on to set (possibly empty) `alert_rules` in the relation data. This works because its `_resolve_dir_against_charm_path` doesn't align with the rest of the libs, and it doesn't throw an exception. Instead, it sets it to `os.cwd()` and returns that.
* Prometheus will simply [explode with `InvalidAlertRulePathError` when is refresh is attempted where the path doesn't exist](https://github.com/canonical/prometheus-k8s-operator/blob/60d227d97700d4c0e6c36dae7838b4c74d78f197/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L795). There's no `try/except` around that call.  So, yes, it needs unit tests, because "update the config to some unresolvable dir" is currently -> `Exception`, not "empty relation data"
- That's assuming it even gets that far, because `_resolve_dir_against_charm_path` will throw an exception in the constructor, and...

Only Loki would actually "clear" the rules, and that's probably only because nobody has synced up`_resolve_dir_against_charm_path` yet, otherwise the pattern from Prometheus would lead to:

```python
try:
  alert_rules_path = _resolve_...
except Invalid...
  log(...)

# alert_path is unbound
# NameError: name 'alert_rules_path'  is not defined
self._alert_rules_path = alert_path
```

`cos-config` is likely to run face first into "an exception was thrown". We should handle these errors better in the Prometheus library in general, but as it stands, library behavior is not the same across our bundle.

Calling the method `reinitialize()` or `rebuild()` or adding `purge=True` (which `cos-config` can freely call) is an easy guard rail since `cos-config` will go into `BlockedStatus` anyway.

**ESPECIALLY** considering that we are exposing a public method in the API for libraries which does not advertise itself as potentially destructive through any mechanism (method name, optarg even if the default is `purge_existing=True`) to enable the use case of a single charm, one of our own, which could already easily "know" whether or not it even needs that flag by checking the validity of the paths before it blindly passes them to library constructors, could easily wipe the relation databags from the charm code, etc.

Adding dangerous default behavior to libraries for a single "user" is robbing Peter to pay Paul.

rbarry82: As an additional note, this is a great example of the kind of "breaking" backwards compatibility we talk about in relation to Hyrum's law.

Charmhelpers [has Ansible support](https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/ansible/__init__.py), we literally [gave a talk on it](https://docs.google.com/presentation/d/1e0jKfeQmCRLasFGePGFZb_6dg2IDgaY_gW6O2iICHJM/htmlpresent), and [the ElasticSearch charm uses it](https://git.launchpad.net/charm-elasticsearch/tree/).

"Hey, we changed how things work when you rebuild your charm and don't change the args ala [the change to `PrometheusRulesProvider`](https://github.com/canonical/prometheus-k8s-operator/pull/209)" is certainly what I would consider a breaking change, since it's extraordinarily likely considering that `juju show-unit ...` allows for easy introspection, and `juju relation-set` would trivially allow someone to go re-implement their own version of `cos-config` which sets the relation data for alert rules by pulling from some K/V store based on the AWS region or... whatever.

Blowing away their rules (or similar) when they set some relation data and call `.refresh()` (or wait for an event) will be breaking changes, and bugs. We can say "don't do that" or "this is the supported use case" or something similar, but it is what it is. 

This kind of behavior change is "opt-in" behavior, and `cos-config` "knows" that it wants to opt into it.

sed-i: To summarize, iiuc, you're proposing:

1. introduce the "last known good configuration" concept into (our) charms (cf. [operator/649](https://github.com/canonical/operator/issues/649))
2. allow cos-config to override the last-known-good-configuration with a `--force` flag to `reload_dashboards` and `reload_alerts` 
3. (changing the status message is an easy agreeable)

Correct?

mmanciop: The "Last known configuration" concept is not something some charms should introduce. Please implement https://github.com/canonical/grafana-k8s-operator/pull/53#discussion_r791452643 and I take the responsibility on me :-)

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -590,6 +592,36 @@ def _update_all_dashboards_from_dir(self, _: HookEvent) -> None:
...
+            stored_dashboard_templates = self._stored.dashboard_templates
+
+            for dashboard_id in list(stored_dashboard_templates.keys()):
+                if dashboard_id.startswith("file:"):
+                    del stored_dashboard_templates[dashboard_id]
+            self._stored.dashboard_templates = stored_dashboard_templates
+
+            # With al of the file-based dashboards cleared out, force a refresh
```
sed-i, line 623:
```suggestion
            # With all of the file-based dashboards cleared out, force a refresh
```

sed-i (approved):

sed-i (approved):

PR #54, opened 2022-01-26
Be more resilient to timing with the unit address

rbarry82:
It's possible for network.get_binding to return `None` if an
address is not set, which results in a source address which is
something like `None:9090` when interpolated.

If the address can't be retrieved, don't set it in the unit data
bucket. The refresh event (by default, `pebble_ready`) will catch
this. This helps us avoid confusing/breaking the Grafana database
by submitting multiple provisioning events which can include
adding then immediately removing a source from `deleteDatasources`
if a relation is rapidly broken/re-joined.

While we're at it, hash the values by pulling from the container
and rework the charm tests since the Pebble harness works

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -5,16 +5,18 @@
 
 import json
 import logging
-from typing import Dict, List, Optional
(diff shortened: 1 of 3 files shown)


sed-i (approved):
lgtm

PR #58, opened 2022-02-03
Allow serving Grafana from a subpath

rbarry82:


src/charm.py
```diff
@@ -463,10 +467,56 @@ def restart_grafana(self) -> None:
...
+        elif parts.path.startswith("/"):
+            ret["scheme"] = "http"
+            ret["host"] = "0.0.0.0"
+            ret["port"] = "3000"
+            ret["path"] = parts.path
+        else:
+            # This could really be anything!
+            logger.warning(
+                "Could not determine web_external_url for Grafana. Please "
+                "use a fully-qualified path or a bare subpath"
```
sed-i, line 490: Add this to the docstring?

src/charm.py
```diff
@@ -463,10 +467,56 @@ def restart_grafana(self) -> None:
                 "not exist or is not responsive"
             )
 
+    def _parse_grafana_path(self, parts: ParseResult) -> dict:
+        """Convert web_external_url into a usable path."""
```
sed-i, line 471: Can you add to `config.yaml`'s description a few examples for valid `web_external_url`? For example, is `juju config grafana web_external_url="/grafana"` valid?

sed-i (approved):

rbarry82:
I'm gonna hold off on changes until it's confirmed working, then will update the docs

src/charm.py
```diff
@@ -463,10 +467,56 @@ def restart_grafana(self) -> None:
...
+            ret["host"] = "0.0.0.0"
+            ret["port"] = "3000"
+            ret["path"] = parts.path
+        else:
+            # This could really be anything!
+            logger.warning(
+                "Could not determine web_external_url for Grafana. Please "
+                "use a fully-qualified path or a bare subpath"
+            )
+
```
simskij, line 492:
```suggestion
        if not parts.scheme and not parts.path.startswith("/"):
            # This could really be anything!
            logger.warning(
                "Could not determine web_external_url for Grafana. Please "
                "use a fully-qualified path or a bare subpath"
            )
            return {}
            
        return {
            "scheme": parts.scheme or "http",
            "host": "0.0.0.0",
            "port": parts.netloc.split(":")[1] if ":" in parts.netloc else "3000",
            "path": parts.path,
        }
```

This should be fairly equivalent, but maybe a bit easier to reason about. Feel free to just reject it if you prefer what you already have.

simskij (approved):

src/charm.py
rbarry82: I like this --- admittedly, I'm more or less waiting for confirmation of "does this work"?

I should setup a microk8s instance with an ingress and test it, but the patch is up to unblock internal testing, and it's that useful to any consumers until we resolve the ingress story

tests/unit/test_charm.py
```diff
@@ -175,3 +175,13 @@ def can_get_password(self):
 
         # Harness doesn't quite support actions yet...
         self.assertTrue(re.match(r"[A-Za-z0-9]{12}", self.harness.charm._get_admin_password()))
+
+    def test_config_is_updated_with_subpath(self):
+        self.harness.set_leader(True)
+
+        self.harness.update_config({"web_external_url": "/grafana"})
+        self.harness.charm.on.config_changed.emit()
```
sed-i, line 183: config changed is emitted by update config
https://github.com/canonical/operator/blob/7ff4963d23a7f101330bbeef83da72bd1bbfe305/ops/testing.py#L881

```suggestion
```

tests/unit/test_charm.py
```diff
@@ -175,3 +175,13 @@ def can_get_password(self):
...
+
+        services = self.harness.charm.container.get_plan().services["grafana"].to_dict()
+        self.assertIn("GF_SERVER_SERVE_FROM_SUB_PATH", services["environment"].keys())
+        self.assertIn("/grafana", services["environment"]["GF_SERVER_ROOT_URL"])
```
sed-i, line 187:
```suggestion
        self.assertTrue(services["environment"]["GF_SERVER_ROOT_URL"].endswith("/grafana"))
```

sed-i (dismissed):

src/charm.py
```diff
@@ -449,11 +453,7 @@ def restart_grafana(self) -> None:
             plan = self.container.get_plan()
             if plan.services != layer.services:
                 self.container.add_layer(self.name, layer, combine=True)
-
-                if self.container.get_service(self.name).is_running():
-                    self.container.stop(self.name)
-
-                self.container.start(self.name)
+                self.container.replan()
```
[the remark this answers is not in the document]
rbarry82: There are no plans to change this at any point

src/charm.py
```diff
@@ -192,6 +193,9 @@ def _configure(self) -> None:
 
             restart = True
 
+        if self.container.get_plan().services != self._build_layer().services:
```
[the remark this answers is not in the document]
rbarry82: No. Immediately above this, during `_update_...`, files are pushed. It should always be running by this point.

As much as "should" is a terrible word, significant changes to the logic of the code should not be made to enable an internal, testing-only workflow, no matter whether they look safe or not.

sed-i (approved):
lgtm, apart from the minor lint and static :)

rbarry82:
Sometimes words cannot express how profoundly irritating it is to have `mypy` succeed locally and fail in CI despite the use of as much abstraction as we can get

sed-i (approved):

src/charm.py
```diff
@@ -463,10 +466,50 @@ def restart_grafana(self) -> None:
...
+            logger.warning(
+                "Could not determine web_external_url for Grafana. Please "
+                "use a fully-qualified path or a bare subpath"
+            )
+            return {}
+
+        return {
+            "scheme": parts.scheme or "http",
+            "host": "0.0.0.0",
+            "port": parts.netloc.split(":")[1] if ":" in parts.netloc else "3000",
```
Abuelodelanada, line 484: Why not using the constant `PORT` instead of "3000"?

Abuelodelanada (approved):

PR #59, opened 2022-02-07
minor doc fixes

sed-i:
A few fixes I encountered while working on the load test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,8 @@ View the dashboard in a browser:
 2. Navigate to `http://IP_ADDRESS:3000`
 3. Log in with the credentials username=admin, password=<from run-action> (these credentials are configurable at deploy time)
 
-To manually set the admin password:
-```bash
-microk8s.kubectl exec -n MODEL_NAME grafana-0 -c grafana -- grafana-cli admin reset-admin-password new_password
-```
+To manually set the admin password, see the
+[official docs](https://grafana.com/docs/grafana/latest/administration/cli/#reset-admin-password).
 
 Additionally, Grafana can be accessed via the Kubernetes service matching the Juju application name in the namespace matching the Juju model's name.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ juju run-action grafana-k8s/0 get-admin-password --wait
 View the dashboard in a browser:
 1. `juju status` to check the IP of the running Grafana application
 2. Navigate to `http://IP_ADDRESS:3000`
-3. Log in with the credentials username=admin, password=<from run-action> (these credentials are configurable at deploy time)
+3. Log in with the username `admin`, and the password you got from the `get-admin-password` action.
 
 To manually set the admin password, see the
 [official docs](https://grafana.com/docs/grafana/latest/administration/cli/#reset-admin-password).
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Grafana Operator provides a data visualization solution using [Grafana](http
 observability toolkit.
 
 Grafana allows you to query, visualize, alert on, and visualize metrics from mixed datasources in configurable
-dashboards for observability. This repository contains a [Juju](https://jaas.ai/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster.
+dashboards for observability. This repository contains a [Juju](https://juju.is/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster.
 
 The grafana-k8s charm provides an interface which can ingest data from a wide array of data sources, with Prometheus as a common input, then presents that data on configurable dashboards. It is the primary user-facing entrypoint for the Canonical Observability Stack Lite. See the [COS Lite Bundle](https://charmhub.io/cos-lite) for more information.
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Grafana Operator provides a data visualization solution using [Grafana](http
 observability toolkit.
 
 Grafana allows you to query, visualize, alert on, and visualize metrics from mixed datasources in configurable
-dashboards for observability. This repository contains a [Juju](https://jaas.ai/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster. 
+dashboards for observability. This repository contains a [Juju](https://jaas.ai/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster.
 
 The grafana-k8s charm provides an interface which can ingest data from a wide array of data sources, with Prometheus as a common input, then presents that data on configurable dashboards. It is the primary user-facing entrypoint for the Canonical Observability Stack Lite. See the [COS Lite Bundle](https://charmhub.io/cos-lite) for more information.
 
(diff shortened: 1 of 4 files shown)


rbarry82 (dismissed):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -30,9 +34,19 @@
     path/to/src/grafana_dashboards/*.tmpl
 
 Where the `*.tmpl` files are Grafana dashboard JSON data either from the
-Grafana marketplace, or directly exported from a a Grafana instance.
+Grafana marketplace, or directly exported from a Grafana instance.
+Dashboards obtain via export or via the marketplace would need to be renamed to have the `*.tmpl`
+suffix, otherwise they will not be read.
+Each `*.tmpl` file should have a distinct set of "uid" and "version" - the grafana web ui will not
+show dashboards with colliding uid and version.
```
sed-i, line 41: @rbarry82 - replying here because [your comment](https://github.com/canonical/grafana-k8s-operator/commit/559c7bab10e3b53ce564bed392d76664c6fe18bc#commitcomment-66304608) is attached to a commit and cannot be replied to.

> This is exactly the kind of thing we should not document.
>
> The behavior is in no way specific to charms, is [clearly documented](https://grafana.com/tutorials/provision-dashboards-and-data-sources/) by Grafana themselves, and could change in any given release, leaving our docs out of line with actual behavior.
>
> I realize that it happened to you, but that was an unusual workflow for testing which users are unlikely to ever encounter. If they do, googling the error will find Grafana’s docs before ours, and cleanly resolve it.

I am still finding my way in the darkness so thanks for pointing to the grafana doc.
That doc mentions `id`, I guess they mean the `uid`?

On one hand I agree with you; on the other hand bringing this information closer to where it is consumed is probably a good thing. Many things can change and leave our docs out of date, this is not unique to this line here. Unless uid and version are still volatile..?

The user is already asked to use `prometheusds`, rename the file to `.tmpl` and possibly remove `current {` sections. It is not unthinkable that the use would question the charm before grafana if something is not working, and this comment might help?

Anyway, not sure what's better here.

README.md
```diff
@@ -6,7 +6,7 @@ The Grafana Operator provides a data visualization solution using [Grafana](http
 observability toolkit.
 
 Grafana allows you to query, visualize, alert on, and visualize metrics from mixed datasources in configurable
-dashboards for observability. This repository contains a [Juju](https://jaas.ai/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster. 
+dashboards for observability. This repository contains a [Juju](https://jaas.ai/) Charm for deploying the visualization component of Grafana in a Kubernetes cluster.
```
mmanciop, line 9: The link should go to juju.is, not jaas.ai

README.md
```diff
@@ -35,11 +35,16 @@ This unit and its IP address can be retrieved using the `juju status` command.
...
 ```bash
 juju run-action grafana-k8s/0 get-admin-password --wait
+```
+
 View the dashboard in a browser:
-1. `juju status` to check the IP of the of the running Grafana application
+1. `juju status` to check the IP of the running Grafana application
 2. Navigate to `http://IP_ADDRESS:3000`
-3. Log in with the default credentials username=admin, password=admin (these credentials are configurable at deploy time)
+3. Log in with the credentials username=admin, password=<from run-action> (these credentials are configurable at deploy time)
```
mmanciop, line 43: I don’t think this is understandable. Please add a section about retrieving the admin pwd if we do not have one, and ljnk it from here

lib/charms/grafana_k8s/v0/grafana_dashboard.py
mmanciop: Technically nothing prevents us to automatically detect the variable used by the Prometheus datasource (assuming there is only one), and adjust out logic. Care to open an issue about it for Ryan?

rbarry82: Yes, they mean the uid, probably.

I think. the difference to me in this particular string is that the set of users who would follow this workflow is essentially "this team" (and maybe some internal to Grafana). End users who follow Grafana's docs (or ours when they are more fleshed out) about how to export dashboards for consumption will not, nor will users who get them from the marketplace, or anyone who uses JSONNET.

The only scenario in which this is a "real" possibility is a user who decides to copy and edit a template by hand as raw JSON, or one who's making nearly-identical dashboards (which is rare). 

There are countless "don't do this" things we could document, but do not. Documentation is also a tradeoff between completeness and complexity. "We encountered this edge case in testing which had unexpected behavior, and 999/1000 users will never see it, but we'll document it anyway" rapidly becomes something of questionable utility which is hard for users to navigate.

Instead, documenting "here is how to export a dashboard" reduces the chances of this in the same way, while also offering benefit.

@mmanciop there was a reason we did not, but let me walk through the code again, because enough things have changed since the original "three legged table" implementation that it may be plausible now

sed-i: I think I encountered this in a reasonable workflow, which I wouldn't call manual copy-paste; in plain text the dashboards are quite different (big diff), don't know yet how lzma fits here.

1. Open a pre-provisioned dashboard.
2. Modify it, try to save, grafana prevents and offers to export as JSON, so save as JSON on my machine.
3. Repeat the above.

As a results I get exported dashboards with the same uid and version.

README.md
sed-i: Thanks! Fixed in next commit.

README.md
sed-i: The action is described in line 37.

balbirthomas (dismissed):
This does not pertain to this PR but I am surprised that there is an action to get the admin password but no corresponding action to set the admin password and instead the user has to resort to a `kubectl` command . @rbarry82 shall we create a ticket for this ?

balbirthomas:
> > This does not pertain to this PR but I am surprised that there is an action to get the admin password but no corresponding action to set the admin password and instead the user has to resort to a `kubectl` command . @rbarry82 shall we create a ticket for this ?
> 
> @balbirthomas: @rbarry82 already kindly explained to me that if we set it, then it would be available in plaintext in the model, which could be a problem.

Thanks for clarifying. Good point.

rbarry82:
I missed that in the review. @sed-i we should **not** be documenting “please `kubectl …`”. 

The user story to set it is “retrieve the password with the action, use the Grafana UI”. 

Using `kubectl` is something we do in our own tests, and only because the other machine can’t talk to Juju. Similar to the dashboard uid thing, just because we encountered a need for this in our own hacked up internal workflows doesn’t mean it belongs in user-facing documentation.

simskij:
I do **not** want us to add the ability to set the password. It should be a one-off initial password, that is randomly generated, by design.

> The user story to set it is “retrieve the password with the action, use the Grafana UI”.

...sums it up very well.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: LZMA doesn’t affect this. 

The key thing is that the `uid` field in the JSON is identical. Not anything Python. 

This workflow is rare. It’s so rare that  Grafana didn’t log anything above debug. 

Even if it let you save, the uid would have been identical, just a different version set, and it would have been the same problem. The workflow used to create these is an anti pattern

simskij:
> I agree it does not belong in the actions.yaml but I would certainly keep it in the other docs - aren't we being a dumb-down nanny for the user too much here? Honest question.

**If** we want to mention it in the docs, I think we should link to [the relevant section in the grafana docs](https://grafana.com/docs/grafana/latest/administration/cli/#reset-admin-password) rather than rehashing it ourselves. With that said, I'd personally rather not have it documented as it will give off the idea that we encourage doing it, which I think we shouldn't.

rbarry82:
Or, put differently, we are documenting how to use _Charmed_ Grafana, not Grafana. We cannot (and should not) try to duplicate even a significant subset of Grafana/Loki/Prometheus/whatever. 

We are documenting how to integrate with the charmed versions, within the Juju ecosystem.

README.md
```diff
@@ -42,10 +42,8 @@ View the dashboard in a browser:
 2. Navigate to `http://IP_ADDRESS:3000`
 3. Log in with the credentials username=admin, password=<from run-action> (these credentials are configurable at deploy time)
```
simskij, line 43:
```suggestion
3. Log in with the username `admin`, and the password you got from the `get-admin-password` action.
```

simskij (dismissed):

balbirthomas (approved):

rbarry82 (approved):

PR #63, opened 2022-02-10
Do a lot more work on dashboard templates

### LINKED ISSUE #60: json error in provided dashboard sends the charm into error state
I had a stray comma in the json dashboard provided, which resulted in:

```
unit-grafana-0: 02:40:24 ERROR unit.grafana/0.juju-log grafana-dashboard:9: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 657, in <module>
    main(GrafanaCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 426, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 276, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 736, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/grafana_k8s/v0/grafana_dashboard.py", line 938, in _on_grafana_dashboard_relation_changed
    self._render_dashboards_and_emit_event(event.relation)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/grafana_k8s/v0/grafana_dashboard.py", line 1020, in _render_dashboards_and_emit_event
    content = _encode_dashboard_content(_inject_dashboard_dropdowns(content))
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/grafana_k8s/v0/grafana_dashboard.py", line 522, in _inject_dashboard_dropdowns
    dict_content = json.loads(content)
  File "/usr/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.8/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
```

Bug reference: OPENG-616

rbarry82:
Return an appropriate error to the Provider if the JSON is bad.

If there are no template variables, assume any datasources are
prometheus (impossible to tell from `panel` objects), and replace
their values with `${prometheusds}`

If there **are** template variables, sift through them. If they
are known types for Loki or Prometheus, replace them by name in the
templates, then go back and clean them up from template variables.

Add tests for all of this, obviously.

Closes #60

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -526,11 +526,16 @@ def _convert_dashboard_fields(content: str) -> str:
     dict_content = json.loads(content)
     datasources = {}
     existing_templates = False
-    if "templating" not in content:
+
+    # If no existing template variables exist, just insert our own
+    if "templating" not in dict_content:
         dict_content["templating"] = {"list": [d for d in TEMPLATE_DROPDOWNS]}
     else:
+        # Otherwise, set a flag so we can go back later
         existing_templates = True
         for maybe in dict_content["templating"]["list"]:
+            # Build a list of `datasource_name`: `datasource_type` mappings
+            # The "query" field is actually "prometheus", "loki", "influxdb", etc
             if "type" in maybe and maybe["type"] == "datasource":
                 datasources[maybe["name"]] = maybe["query"]
 
@@ -560,6 +565,12 @@ def _replace_template_fields(
     if datasources or not existing_templates:
         panels = dict_content["panels"]
 
+        # Go through all of the panels. If they have a datasource set, AND it's one
+        # that we can convert to ${lokids} or ${prometheusds}, by stripping off the
+        # ${} templating and comparing the name to the list we built, replace it,
+        # otherwise, leave it alone.
+        #
+        # COS only knows about Prometheus and Loki.
         for panel in panels:
             if "datasource" not in panel:
                 continue
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -200,7 +200,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 10
 
 logger = logging.getLogger(__name__)
 
@@ -517,17 +517,75 @@ def _decode_dashboard_content(encoded_content: str) -> str:
     return lzma.decompress(base64.b64decode(encoded_content.encode("utf-8"))).decode()
 
 
-def _inject_dashboard_dropdowns(content: str) -> str:
-    """Make sure dropdowns are present for Juju topology."""
+def _convert_dashboard_fields(content: str) -> str:
+    """Make sure values are present for Juju topology.
+
+    Inserts Juju topology variables and selectors into the template, as well as
+    a variable for Prometheus.
+    """
     dict_content = json.loads(content)
+    datasources = {}
+    existing_templates = False
     if "templating" not in content:
         dict_content["templating"] = {"list": [d for d in TEMPLATE_DROPDOWNS]}
     else:
+        existing_templates = True
+        for maybe in dict_content["templating"]["list"]:
+            if "type" in maybe and maybe["type"] == "datasource":
+                datasources[maybe["name"]] = maybe["query"]
+
+        # Put our own variables in the template
         for d in TEMPLATE_DROPDOWNS:
             if d not in dict_content["templating"]["list"]:
                 dict_content["templating"]["list"].insert(0, d)
 
+    dict_content = _replace_template_fields(dict_content, datasources, existing_templates)
+
     return json.dumps(dict_content)
+
+
+def _replace_template_fields(
+    dict_content: dict, datasources: dict, existing_templates: bool
+) -> dict:
+    """Make templated fields get cleaned up afterwards.
+
+    If existing datasource variables are present, try to substitute them, otherwise
+    assume they are all for Prometheus and put the prometheus variable there.
+    """
+    replacements = {"loki": "${lokids}", "prometheus": "${prometheusds}"}
+    used_replacements = []
+
+    # If any existing datasources match types we know, or we didn't find
+    # any templating variables at all, template them.
+    if datasources or not existing_templates:
+        panels = dict_content["panels"]
+
+        for panel in panels:
+            if "datasource" not in panel:
+                continue
+            if not existing_templates:
+                panel["datasource"] = "${prometheusds}"
+            else:
+                # Strip out variable characters and maybe braces
+                ds = re.sub(r"(\$|\{|\})", "", panel["datasource"])
+                replacement = replacements.get(datasources[ds], "")
+                if replacement:
+                    used_replacements.append(ds)
+                panel["datasource"] = replacement or panel["datasource"]
+
+        # Put our substitutions back
+        dict_content["panels"] = panels
+
+    # Finally, go back and pop off the templates we stubbed out
+    deletions = []
+    for tmpl in dict_content["templating"]["list"]:
+        if tmpl["name"] and tmpl["name"] in used_replacements:
+            deletions.append(tmpl)
+
+    for d in deletions:
+        dict_content["templating"]["list"].remove(d)
+
+    return dict_content
 
 
 def _type_convert_stored(obj):
@@ -1017,7 +1075,10 @@ class GrafanaDashboardConsumer(Object):
             error = None
             try:
                 content = Template(decoded_content).render()
-                content = _encode_dashboard_content(_inject_dashboard_dropdowns(content))
+                content = _encode_dashboard_content(_convert_dashboard_fields(content))
+            except json.JSONDecodeError as e:
+                error = str(e.msg)
+                relation_has_invalid_dashboards = True
             except TemplateSyntaxError as e:
                 error = str(e)
                 relation_has_invalid_dashboards = True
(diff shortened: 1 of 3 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
+def _convert_dashboard_fields(content: str) -> str:
+    """Make sure values are present for Juju topology.
+
+    Inserts Juju topology variables and selectors into the template, as well as
+    a variable for Prometheus.
+    """
     dict_content = json.loads(content)
+    datasources = {}
+    existing_templates = False
     if "templating" not in content:
```
sed-i, line 529: Searching a substring like this seems wrong. I guess this is a typo and should be
```suggestion
    if "templating" not in dict_content.keys():
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
         dict_content["templating"] = {"list": [d for d in TEMPLATE_DROPDOWNS]}
     else:
+        existing_templates = True
+        for maybe in dict_content["templating"]["list"]:
```
sed-i, line 533: Potential KeyError shouldn't send the charm into error state :)

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
+            if "type" in maybe and maybe["type"] == "datasource":
+                datasources[maybe["name"]] = maybe["query"]
```
sed-i, line 535: This line with all the "maybe"s... :D Maybe a better name for maybe, maybe not perhaps?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
     return json.dumps(dict_content)
 
 
+def _replace_template_fields(
+    dict_content: dict, datasources: dict, existing_templates: bool
+) -> dict:
+    """Make templated fields get cleaned up afterwards.
+
+    If existing datasource variables are present, try to substitute them, otherwise
+    assume they are all for Prometheus and put the prometheus variable there.
```
sed-i, line 553: Can you please document the purpose of `existing_templates`? I don't get it

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
+    # If any existing datasources match types we know, or we didn't find
+    # any templating variables at all, template them.
+    if datasources or not existing_templates:
+        panels = dict_content["panels"]
+
+        for panel in panels:
+            if "datasource" not in panel:
+                continue
+            if not existing_templates:
+                panel["datasource"] = "${prometheusds}"
```
sed-i, line 567: This seems a bit aggressive:
1. Shouldn't users be able to use dashboard with hardcoded literals for datasources?
2. Wouldn't this make the grafana charm very much a cos-grafana charm? i.e. absolutely not reusable outside of COS.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -517,19 +517,77 @@ def _decode_dashboard_content(encoded_content: str) -> str:
...
+            else:
+                # Strip out variable characters and maybe braces
+                ds = re.sub(r"(\$|\{|\})", "", panel["datasource"])
+                replacement = replacements.get(datasources[ds], "")
```
sed-i, line 571: I don't understand this. Is this for supporting template variables `loki` and `prometheus` in addition to `lokids` and `prometheusds`?

tests/unit/test_dashboard_consumer.py
```diff
@@ -171,3 +266,82 @@ def test_consumer_error_on_bad_template(self):
...
             ],
         )
+
+    def test_consumer_error_on_bad_json(self):
+        self.assertEqual(len(self.harness.charm.grafana_consumer._stored.dashboards), 0)
+        self.assertEqual(self.harness.charm._stored.dashboard_events, 0)
+        rels = self.setup_charm_relations()
+        self.assertEqual(self.harness.charm._stored.dashboard_events, 1)
+
+        bad_data = {
```
sed-i, line 276: Comment what's bad about it?
Can we have more variations of bad that are not json errors?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: It should not be, and this isn’t a typo. It hasn’t changed in a long time.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: `list` is a required subkey in the Grafana dashboard spec.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: “Maybe” is a common enough monadic pattern that there are 5+ libraries in Python to implement it, PEP 505 has a “maybe dot” operator, and is we could rely on a pep for structural pattern matching, this would use that. 

The naming is obvious to anyone who has ever used functors, ML, Haskell, Rust, Scala, etc. It was chosen on purpose.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: It’s passed in from immediately above, on the same screen of code, in a private method. It probably doesn’t need docs. 

It’s set of the template we get already has template variables

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: This is used as a join key. We set it above, and say “if we find a templated variable, store the name and type”. Here, we say “when we look up the type, see if we have a matching substitution “

This is to replace any Loki/Prometheus datasource which was templated and all of its occurrences, whether `$foo` or `${foo}` in the panels with `${prometheusds}` or `${lokids}`

tests/unit/test_dashboard_consumer.py
rbarry82: It’s bad because it throws an exception. 

There is another which fails jinja template rendering immediately above this.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: 1. Not on import. “We want to ensure that imported dashboards automatically work with juju-ized COS Prometheus” and “you can point to anything you want, including Prometheus datasources not related to Grafana” are orthogonal choices. It’s a better trade off to make the 95% use case easy than the 5%. 

2. No. Again, only on import through the library, but we would expect that to _be_ a “cos” dashboard anyway. 

There is not a reasonable balance between flexibility and making it easy with topology and relations, and we don’t want everyone to need to go through the exercise you went through yesterday just to make a dashboard work seamlessly. 

The alternative is adding a constellation of optargs to support niche use cases. We can cross that bridge later if there’s any need

rbarry82:
Note: I resolved most of these because it’s unlikely that any amount of comments will explain “here’s what a dashboard looks like” to demonstrate why this is happening, similar to the Aggregator. 

Manipulating the dashboard is necessarily dense code. The test cases are an excellent demonstration of input->output which offer better explanations than any comment would unless an entire partial state tree was left in the comments, for a private method.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
dstathis: I agree that is probably a bit too COS specific. Perhaps the lib constructor can have an arg like `covert_for_cos=True`?

rbarry82: This would always be set, since it’s from the consumer side of the library.

Abuelodelanada (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1017,7 +1086,10 @@ def _render_dashboards_and_emit_event(self, relation: Relation) -> None:
...
             try:
                 content = Template(decoded_content).render()
-                content = _encode_dashboard_content(_inject_dashboard_dropdowns(content))
+                content = _encode_dashboard_content(_convert_dashboard_fields(content))
+            except json.JSONDecodeError as e:
+                error = str(e.msg)
+                relation_has_invalid_dashboards = True
             except TemplateSyntaxError as e:
                 error = str(e)
                 relation_has_invalid_dashboards = True
```
dstathis, line 1095: This should be `except (json.JSONDecodeError, TemplateSyntaxError) as e:`

dstathis (approved):

PR #65, opened 2022-02-15
Move away from StoredState for Grafana.*Consumer and charm app/unit data

rbarry82:
We can't rely on a peer relation in the aggregator or a provider
charms, so we're left with StoredState there, but get rid of it
from Grafana and the associated Consumer libraries.

Update the tests to match. Since JSON doesn't have a literal
int type, make sure that the relation keys in the tests also are
not bare ints, and that we convert from Stored*

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -1055,7 +1055,7 @@ class GrafanaDashboardConsumer(Object):
             )
 
             for relation in relations:
-                self._render_dashboards_and_signal_changed(relation)
+                self._render_dashboards_and_signal_changed(relation)  # type: ignore
 
         if changes:
             self.on.dashboards_changed.emit()
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -1090,7 +1090,7 @@ class GrafanaDashboardConsumer(Object):
             self.on.dashboards_changed.emit()
         else:
             stored_data = rendered_dashboards
-            currently_stored_data = self._get_stored_dashboards(relation.id) or {}  # type: ignore
+            currently_stored_data = self._get_stored_dashboards(relation.id)  # type: ignore
 
             coerced_data = (
                 _type_convert_stored(currently_stored_data) if currently_stored_data else {}
@@ -1099,7 +1099,7 @@ class GrafanaDashboardConsumer(Object):
             if not coerced_data == stored_data:
                 stored_dashboards = self.get_peer_data("dashboards")
                 stored_dashboards[relation.id] = stored_data
-                self.set_peer_data({"dashboards": stored_dashboards})
+                self.set_peer_data("dashboards", stored_dashboards)
                 self.on.dashboards_changed.emit()
 
     def _remove_all_dashboards_for_relation(self, relation: Relation) -> None:
@@ -1107,7 +1107,7 @@ class GrafanaDashboardConsumer(Object):
         if self._get_stored_dashboards(relation.id):
             stored_dashboards = self.get_peer_data("dashboards")
             stored_dashboards.pop(relation.id)
-            self.set_peer_data({"dashboards": stored_dashboards})
+            self.set_peer_data("dashboards", stored_dashboards)
             self.on.dashboards_changed.emit()
 
     def _to_external_object(self, relation_id, dashboard):
@@ -1144,12 +1144,11 @@ class GrafanaDashboardConsumer(Object):
         data = {"dashboards": {}}  # type: ignore
         for k, v in data.items():
             if not self.get_peer_data(k):
-                self.set_peer_data({k: v})
+                self.set_peer_data(k, v)
 
-    def set_peer_data(self, data: dict) -> None:
+    def set_peer_data(self, key: str, data: Any) -> None:
         """Put information into the peer data bucket instead of `StoredState`."""
-        for k, v in data.items():
-            self._charm.peers.data[self._charm.app][k] = json.dumps(v)  # type: ignore
+        self._charm.peers.data[self._charm.app][key] = json.dumps(data)  # type: ignore
 
     def get_peer_data(self, key: str) -> Any:
         """Put information into the peer data bucket instead of `StoredState`."""
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -926,6 +926,10 @@ class GrafanaDashboardConsumer(Object):
             self._charm.on[self._relation_name].relation_broken,
             self._on_grafana_dashboard_relation_broken,
         )
+        self.framework.observe(
+            self._charm.on[DEFAULT_PEER_NAME].relation_changed,
+            self._on_grafana_peer_changed,
+        )
 
     def get_dashboards_from_relation(self, relation_id: int) -> List:
         """Get a list of known dashboards for one instance of the monitored relation.
@@ -957,6 +961,12 @@ class GrafanaDashboardConsumer(Object):
 
         if changes:
             self.on.dashboards_changed.emit()
+
+    def _on_grafana_peer_changed(self, _: RelationChangedEvent) -> None:
+        """Emit dashboard events on peer events so secondary charm data updates."""
+        if self._charm.unit.is_leader():
+            return
+        self.on.dashboards_changed.emit()
 
     def update_dashboards(self, relation: Optional[Relation] = None) -> None:
         """Re-establish dashboards on one or more relations.
@@ -1113,7 +1123,7 @@ class GrafanaDashboardConsumer(Object):
         """If an errored dashboard is in stored data, remove it and trigger a deletion."""
         if self._get_stored_dashboards(relation.id):
             stored_dashboards = self.get_peer_data("dashboards")
-            stored_dashboards.pop(relation.id)
+            stored_dashboards.pop(str(relation.id))
             self.set_peer_data("dashboards", stored_dashboards)
             self.on.dashboards_changed.emit()
 
@@ -1144,7 +1154,7 @@ class GrafanaDashboardConsumer(Object):
 
     def _get_stored_dashboards(self, relation_id: int) -> list:
         """Pull stored dashboards out of the peer data bucket."""
-        return self.get_peer_data("dashboards").get(relation_id, {})
+        return self.get_peer_data("dashboards").get(str(relation_id), {})
 
     def _set_default_data(self) -> None:
         """Set defaults if they are not in peer relation data."""
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -951,12 +951,12 @@ class GrafanaDashboardConsumer(Object):
         available in the app's datastore object. The Grafana charm can
         then respond to the event to update its configuration.
         """
-        # TODO Are we sure this is right? It sounds like every Grafana unit
-        # should create files with the dashboards in its container.
-        if not self._charm.unit.is_leader():
-            return
+        changes = False
+        if self._charm.unit.is_leader():
+            changes = self._render_dashboards_and_signal_changed(event.relation)
 
-        self._render_dashboards_and_emit_event(event.relation)
+        if changes:
+            self.on.dashboards_changed.emit()
 
     def update_dashboards(self, relation: Optional[Relation] = None) -> None:
         """Re-establish dashboards on one or more relations.
@@ -969,13 +969,17 @@ class GrafanaDashboardConsumer(Object):
                 updated. If not specified, all relations managed by this
                 :class:`GrafanaDashboardConsumer` will be updated.
         """
-        if not self._charm.unit.is_leader():
-            return
+        changes = False
+        if self._charm.unit.is_leader():
+            relations = (
+                [relation] if relation else self._charm.model.relations[self._relation_name]
+            )
 
-        relations = [relation] if relation else self._charm.model.relations[self._relation_name]
+            for relation in relations:
+                self._render_dashboards_and_signal_changed(relation)
 
-        for relation in relations:
-            self._render_dashboards_and_emit_event(relation)
+        if changes:
+            self.on.dashboards_changed.emit()
 
     def _on_grafana_dashboard_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Update job config when providers depart.
@@ -988,7 +992,7 @@ class GrafanaDashboardConsumer(Object):
 
         self._remove_all_dashboards_for_relation(event.relation)
 
-    def _render_dashboards_and_emit_event(self, relation: Relation) -> None:
+    def _render_dashboards_and_signal_changed(self, relation: Relation) -> bool:
         """Validate a given dashboard.
 
         Verify that the passed dashboard data is able to be found in our list
@@ -997,6 +1001,9 @@ class GrafanaDashboardConsumer(Object):
 
         Args:
             relation: Relation; The relation the dashboard is associated with.
+
+        Returns:
+            a boolean indicating whether an event should be emitted
         """
         other_app = relation.app
 
@@ -1008,7 +1015,7 @@ class GrafanaDashboardConsumer(Object):
                 self._relation_name,
                 str(relation.id),
             )
-            return
+            return False
 
         data = json.loads(raw_data)
 
@@ -1087,7 +1094,7 @@ class GrafanaDashboardConsumer(Object):
             )
 
             # Dropping dashboards for a relation needs to be signalled
-            self.on.dashboards_changed.emit()
+            return True
         else:
             stored_data = rendered_dashboards
             currently_stored_data = self._get_stored_dashboards(relation.id)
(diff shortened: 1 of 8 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1116,13 +1128,34 @@ def dashboards(self) -> List[Dict]:
...
+    def _get_stored_dashboards(self, relation_id: int) -> list:
+        """Pull stored dashboards out of the peer data bucket."""
+        return self.get_peer_data("dashboards").get(relation_id, {})
+
+    def _set_default_data(self) -> None:
+        """Set defaults if they are not in peer relation data."""
+        data = {"dashboards": {}}  # type: ignore
+        for k, v in data.items():
+            if not self.get_peer_data(k):
+                self.set_peer_data({k: v})
```
sed-i, line 1147: This interface seems a bit confusing to me - it is called set, but it is really the equivalent of `dict.update`.
Maybe `self.set_peer_data(k1=v1, k2=v2, ...)`?
This way it would be more symmetric with `get_peer_data`.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -468,7 +472,7 @@ def _on_grafana_source_relation_changed(self, event: CharmEvents) -> None:
             if source:
                 sources[rel.id] = source
 
-        self._stored.sources = sources
+        self.set_peer_data({"sources": sources})
```
sed-i, line 475: Not sure but shouldn't this kind of data go into unit data bag?
Originally it was `_stored`, which is per unit, and now it is per app. Just making sure this is intentional etc.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -541,50 +547,87 @@ def _remove_source_from_datastore(self, event: RelationDepartedEvent) -> None:
...
+        self._set_default_data()
         sources = _type_convert_stored(self._stored.sources)
         for rel_id in sources.keys():
             for i in range(len(sources[rel_id])):
                 sources[rel_id][i].update(
                     {k.replace("-", "_"): v for k, v in sources[rel_id][i].items()}
                 )
 
-        self._stored.sources = sources
+        # If there's stored data, merge it and purge it
```
sed-i, line 589: Would `logger.info` about "data migration" be handy here?

metadata.yaml
```diff
@@ -34,7 +34,7 @@ requires:
     interface: db
     limit: 1
 peers:
-  grafana-peers:
+  grafana:
```
sed-i, line 37: Wouldn't this be confusing in `juju status` output? (the Type: peer column is way on the right)
How about "replicas" or "peers"?

src/charm.py
```diff
@@ -90,9 +93,7 @@ def __init__(self, *args):
         self.grafana_service = Grafana("localhost", PORT)
         self._grafana_config_ini_hash = None
         self._grafana_datasources_hash = None
-        self._stored.set_default(
-            database=dict(), pebble_ready=False, k8s_service_patched=False, admin_password=""
-        )
+        self._stored.set_default(k8s_service_patched=False, admin_password="")
```
sed-i, line 96: iirc we said password goes into rel data as well for now (?)

src/charm.py
```diff
@@ -386,7 +402,7 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
             return
 
         # remove the existing database info from datastore
-        self._stored.database = dict()
+        self.set_peer_data({"database": {}})
```
sed-i, line 405: Shouldn't this action be delegated to charm lib, which in turn would `_set_default_data`?

tests/unit/test_charm.py
```diff
@@ -101,6 +101,7 @@ def setUp(self):
         self.harness = Harness(GrafanaCharm)
         self.addCleanup(self.harness.cleanup)
         self.harness.begin()
+        self.harness.add_relation("grafana", "grafana")
```
sed-i, line 104: I suspect that to truly be a peer relation, the app name here must match metadata, i.e. "grafana-k8s".

tests/unit/test_dashboard_consumer.py
```diff
@@ -86,6 +91,7 @@ def setUp(self):
         self.addCleanup(self.harness.cleanup)
         self.harness.set_leader(True)
         self.harness.begin()
+        self.harness.add_relation("grafana", "grafana")
```
sed-i, line 94: "grafana-k8s"?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1116,13 +1128,34 @@ def dashboards(self) -> List[Dict]:
...
+
+    def set_peer_data(self, data: dict) -> None:
+        """Put information into the peer data bucket instead of `StoredState`."""
+        for k, v in data.items():
+            self._charm.peers.data[self._charm.app][k] = json.dumps(v)  # type: ignore
+
+    def get_peer_data(self, key: str) -> Any:
+        """Put information into the peer data bucket instead of `StoredState`."""
+        data = self._charm.peers.data[self._charm.app].get(key, "")  # type: ignore
```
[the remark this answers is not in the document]
rbarry82: They have the same truth value

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: We cannot rely on unit peer bags for solitary units

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: We don’t lot about the other conversion either. If anything, debug. This is housekeeping which an admin doesn’t need to worry about

metadata.yaml
rbarry82: This matches the practice in other charms

src/charm.py
rbarry82: Let’s not muddy this patch with other issues

src/charm.py
rbarry82: Database doesn’t belong to libs.

tests/unit/test_charm.py
rbarry82: It does not.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: This is more to do with the annoying way Python handles nested dict keys `peer.data[app][type][key]`, and it common in libraries doing the same thing (mostly k/v stores or event streaming)

The “right” way is pickling, but it makes `show-unit` really bad

tests/unit/test_charm.py
sed-i: What do you mean - as long as the relation name is a peer relation, harness doesn't care about the app name?
BTW metadata is `name: grafana-k8s`.

rbarry82: It does not. The signature in the harness is:

```python
add_relation(self, relation_name: str, remote_app: str)
```

As long as the relation name matches ([which is `grafana`](https://github.com/canonical/grafana-k8s-operator/pull/65/files#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R37)), this is fine. The _names_ of the applications in peer data don't actually matter to the harness, even though in practice the odds of having another unit in peer relation with a name which doesn't match is zero

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1116,13 +1128,34 @@ def dashboards(self) -> List[Dict]:
...
+        """Set defaults if they are not in peer relation data."""
+        data = {"dashboards": {}}  # type: ignore
+        for k, v in data.items():
+            if not self.get_peer_data(k):
+                self.set_peer_data({k: v})
+
+    def set_peer_data(self, data: dict) -> None:
+        """Put information into the peer data bucket instead of `StoredState`."""
+        for k, v in data.items():
+            self._charm.peers.data[self._charm.app][k] = json.dumps(v)  # type: ignore
```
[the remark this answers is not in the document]
rbarry82: I tested this quite a bit today, and yes, it's reliable

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: This is updated to be `set_peer_data(key, data: Any)`

Like `StoredState`, the data is mostly set "as a blob", so individual updates don't make much sense.

src/charm.py
```diff
@@ -232,10 +233,24 @@ def _update_grafana_config_ini(self, config: str) -> None:
...
+    def peers(self):
+        """Fetch the peer relation."""
+        return self.model.get_relation(PEER)
+
+    def set_peer_data(self, key: str, data: Any) -> None:
+        """Put information into the peer data bucket instead of `StoredState`."""
+        self.peers.data[self.app][key] = json.dumps(data)
+
+    def get_peer_data(self, key: str) -> Any:
+        """Put information into the peer data bucket instead of `StoredState`."""
```
Abuelodelanada, line 250: Docstring shouldn't be something like:

"""Retrieve information from the peer data bucket instead of `StoredState`."""

Abuelodelanada (approved):

src/charm.py
```diff
@@ -232,10 +233,24 @@ def _update_grafana_config_ini(self, config: str) -> None:
 
     @property
     def has_peers(self) -> bool:
-        """Check whether or nto there are any other Grafanas as peers."""
-        rel = self.model.get_relation(PEER)
+        """Check whether or not there are any other Grafanas as peers."""
+        rel = self.model.get_relation(self.app)
```
sed-i, line 237: Should this be `self.app.name`?
And why not `PEER`?

tests/integration/test_multiple_units.py
```diff
@@ -0,0 +1,147 @@
...
+            grafana_charm, resources=grafana_resources, application_name=grafana_app_name
+        ),
+        ops_test.model.deploy(
+            grafana_tester_charm, resources=tester_resources, application_name=tester_app_name
+        ),
+    )
+    await asyncio.gather(
+        ops_test.model.wait_for_idle(apps=[grafana_app_name], status="active"),
+        ops_test.model.wait_for_idle(apps=[tester_app_name], status="active"),
+    )
```
sed-i, line 48:
```suggestion
    ops_test.model.wait_for_idle(apps=[grafana_app_name, tester_app_name], status="active")
```

tests/integration/test_multiple_units.py
```diff
@@ -0,0 +1,147 @@
...
+            lambda: len(ops_test.model.applications[tester_app_name].units) > 0
+        ),
+    )
+
+    # Idle again to ensure the second unit has an ddress
+    await ops_test.model.wait_for_idle(apps=[grafana_app_name], status="active")
+
+    assert ops_test.model.applications[grafana_app_name].units[0].workload_status == "active"
+    assert ops_test.model.applications[grafana_app_name].units[1].workload_status == "active"
+    assert ops_test.model.applications[tester_app_name].units[0].workload_status == "active"
```
sed-i, line 64: These 3 lines seem to be equivalent to `wait_for_idle`?

tests/unit/test_charm.py
sed-i: ok. I would still choose the less confusing app name though.

src/charm.py
rbarry82: Right, but not "share the password across units in relation data" :)

Some of the Stored* makes sense to keep anyway, so I should rename the PR to reflect the fact that it's _just_ app/unit data being moved.

tests/unit/test_charm.py
rbarry82: For unit testing, it probably doesn't matter, but I picked this because it's also what `mongo` uses and several other charms. It appears to be more or less a de-facto standard.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1169,6 +1218,9 @@ def __init__(
         grafana_relation: str = "downstream-grafana-dashboard",
     ):
         super().__init__(charm, grafana_relation)
+
+        # Reactive charms may be RPC-ish and not leave reliable data around. Keep
+        # StoredState here
```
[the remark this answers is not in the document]
rbarry82: There's not enough information about this. We know that _some_ reactive charms communicate in an RPC-like manner as a conversation to negotiate information. None of the Grafana ones I've found (yet), but better to be safe than sorry, and this is in the Aggregator in any case.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -437,13 +441,19 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
         self._charm = charm
         events = self._charm.on[relation_name]
 
+        # We're stuck with this forever now so upgrades work, or until such point as we can
+        # break compatibility
```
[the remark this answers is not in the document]
rbarry82: Even if there's nothing in `StoredState`, it still needs to be initialized since we can't guarantee which versions will be used.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -458,19 +468,24 @@ def _on_grafana_source_relation_changed(self, event: CharmEvents) -> None:
...
-            if source:
-                sources[rel.id] = source
+            self.set_peer_data("sources", sources)
 
-        self._stored.sources = sources
+        self.on.sources_changed.emit()
 
+    def _on_grafana_peer_changed(self, _: RelationChangedEvent) -> None:
+        """Emit source events on peer events so secondary charm data updates."""
+        if self._charm.unit.is_leader():
```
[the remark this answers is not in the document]
rbarry82: No, there is not. Leaders set the data and won't get this event anyway. This line is possibly superfluous, but harmless.

Integration tests cover this because Harness does not behave nicely with multiple units

sed-i (approved):

tests/integration/test_multiple_units.py
rbarry82: It actually waits for idle just above this, the assertion because we want to really _know_ they're idle instead of hitting the `check_freq` in libjuju for a unit which has been created but not actually doing anything yet.

This is more of a problem when scaling down integration tests, but it doesn't hurt here.

sed-i (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -988,6 +995,10 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
             self._charm.on[self._relation_name].relation_broken,
             self._on_grafana_dashboard_relation_broken,
         )
+        self.framework.observe(
+            self._charm.on[DEFAULT_PEER_NAME].relation_changed,
```
dstathis, line 999: There appears to be no way to set the pear relation name to anything other than default. This seems like an issue.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -437,13 +441,19 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
...
+        # break compatibility
         self._stored.set_default(
             sources=dict(),
             sources_to_delete=set(),
         )
 
         self.framework.observe(events.relation_changed, self._on_grafana_source_relation_changed)
         self.framework.observe(events.relation_departed, self._on_grafana_source_relation_departed)
+        self.framework.observe(
+            self._charm.on[DEFAULT_PEER_NAME].relation_changed,
```
dstathis, line 454: Same thing here about not being able to set peer relation name.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -528,63 +545,104 @@ def _on_grafana_source_relation_departed(self, event: RelationDepartedEvent) ->
...
+        removed_source = stored_sources.pop(str(rel_id), None)
         if removed_source:
             if event.unit:
                 # Remove one unit only
                 dead_unit = [s for s in removed_source if s["unit"] == event.unit.name][0]
                 self._remove_source(dead_unit["source_name"])
 
                 # Re-update the list of stored sources
-                self._stored.sources[rel_id] = [
+                stored_sources[rel_id] = [
```
dstathis, line 575: Is this supposed to be a global?

src/charm.py
```diff
@@ -232,10 +233,24 @@ def _update_grafana_config_ini(self, config: str) -> None:
 
     @property
     def has_peers(self) -> bool:
-        """Check whether or nto there are any other Grafanas as peers."""
+        """Check whether or not there are any other Grafanas as peers."""
         rel = self.model.get_relation(PEER)
         return len(rel.units) > 0 if rel is not None else False
```
dstathis, line 238: I know there wasn't an actual change here but I suspect you may want planned units here

dstathis (changes requested):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: Considering it's only implemented on the consumer side, since we can't rely on peer data buckets necessarily existing on providers at all...

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: It's an instance variable set just above this

src/charm.py
rbarry82: It's actually completely unused since we don't (and will not) set up an external traditional DB

dstathis (approved):

PR #66, opened 2022-02-16
Check the response as well as HTTPError for the password action

### LINKED ISSUE #64: Password shown on action changes after scaling
Hello!

After deploying the cos-lite bundle:
```bash
juju add-model cos
juju deploy cos-lite --channel=edge --trust
```
I can get the dashboard password with:
```
juju run-action grafana get-admin-password
```

If I scale down the grafana app with:
```
juju remove-unit grafana --num-units 1
```
and then scale it back up:
```
juju add-unit grafana --num-units 1
```

Using `get-admin-password` shows a different one, but the dashboard is accessible with the old one.

rbarry82:
If the password is changed, Grafana may not return a 401, but may
include a human readable "invalid username or password" result in
the response. Check this also.

Closes #64

### DIFF
--- a/src/grafana_server.py
+++ b/src/grafana_server.py
@@ -28,7 +28,6 @@ class Grafana:
         Args:
             host: a :str: which indicates the hostname
             port: an :int: to listen on
-
         """
         self.host = host
         self.port = port
@@ -54,8 +53,8 @@ class Grafana:
         headers = urllib3.make_headers(basic_auth="{}:{}".format(username, passwd))
 
         try:
-            self.http.request("GET", url, headers=headers)
-            return False
+            res = self.http.request("GET", url, headers=headers)
+            return True if "invalid username" in res.data.decode("utf8") else False
         except urllib3.exceptions.HTTPError:
             return True
 
--- a/tests/integration/test_get_password.py
+++ b/tests/integration/test_get_password.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+import logging
+from pathlib import Path
+
+import pytest
+import yaml
+from helpers import oci_image
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = "grafana"
+grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+
+
+@pytest.mark.abort_on_fail
+async def test_password_returns_correct_value_after_scaling(ops_test, grafana_charm):
+    """Deploy from charmhub and then upgrade with the charm-under-test."""
+    logger.info("deploying local charm")
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=app_name
+    )
+
+    # set some custom configs to later check they persisted across the test
+
+    action = await ops_test.model.applications[app_name].units[0].run_action("get-admin-password")
+    pw = (await action.wait()).results["admin-password"]
+
+    logger.info("scaling local charm %s to 0 units", grafana_charm)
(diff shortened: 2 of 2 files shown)


src/grafana_server.py
```diff
@@ -54,8 +53,8 @@ def password_has_been_changed(self, username: str, passwd: str) -> bool:
         headers = urllib3.make_headers(basic_auth="{}:{}".format(username, passwd))
 
         try:
-            self.http.request("GET", url, headers=headers)
-            return False
+            res = self.http.request("GET", url, headers=headers)
+            return True if "invalid username" in res.data.decode("utf8") else False
```
sed-i, line 57: What does it return in case the password is valid?
I.e. wouldn't it be less fragile (more future proof) if instead
```suggestion
            return False if res.data.decode("utf8") == "something very predictable" else True
```

sed-i (dismissed):
Add unit test?

src/grafana_server.py
rbarry82: "Invalid username" is very specific.

If the password is valid, it returns a list of orgs, but this is not reliable as a condition depending on how grafana is provisioned -- we cannot look for any specific structure, only a success.

tests/integration/test_get_password.py
```diff
@@ -0,0 +1,43 @@
...
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = "grafana"
+grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+
+
+@pytest.mark.abort_on_fail
+async def test_password_returns_correct_value_after_scaling(ops_test, grafana_charm):
+    """Deploy from charmhub and then upgrade with the charm-under-test."""
```
sed-i, line 22:
```suggestion
    """Deploy the charm under test and then rescale."""
```

tests/integration/test_get_password.py
```diff
@@ -0,0 +1,43 @@
...
+    logger.info("deploying local charm")
+    await ops_test.model.deploy(
+        grafana_charm, resources=grafana_resources, application_name=app_name
+    )
+
+    # set some custom configs to later check they persisted across the test
```
sed-i, line 28:
```suggestion
```

tests/integration/test_get_password.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = "grafana"
+grafana_resources = {"grafana-image": oci_image("./metadata.yaml", "grafana-image")}
+
+
+@pytest.mark.abort_on_fail
+async def test_password_returns_correct_value_after_scaling(ops_test, grafana_charm):
```
sed-i, line 21:
```suggestion
async def test_password_remains_the_same_after_scaling_up_from_0_units(ops_test, grafana_charm):
```

tests/integration/test_get_password.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    # set some custom configs to later check they persisted across the test
+
+    action = await ops_test.model.applications[app_name].units[0].run_action("get-admin-password")
+    pw = (await action.wait()).results["admin-password"]
+
+    logger.info("scaling local charm %s to 0 units", grafana_charm)
+    await ops_test.model.applications[app_name].scale(scale=0)
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) == 0)
```
sed-i, line 35: `wait_for_idle` takes `wait_for_exact_units`
https://github.com/juju/python-libjuju/blob/b9912fce592a40fcda253ed8fde013c1f1607810/juju/model.py#L2567

sed-i (approved):

mmanciop (approved):

PR #67, opened 2022-02-16
Fixed missing datasource for non leader units

balbirthomas:
Non leader units of the Grafana charms do not updated their data
source when new source providers join or depart. This commit
fixes the issue and refactors the unit tests accordingly.


sed-i (commented):
Could you please explain in the PR desc why it is safe to remove the leader guards?
I.e. are there additional leader guards elsewhere that make sure non-leaders do not touch relation data?

PR #68, opened 2022-02-19
Pin markup safe also

rbarry82:
Jinja <3 (via markupsafe) broke today. See
https://github.com/pallets/jinja/issues/1585#issuecomment-1043892628

As they get further way from Python 3.5, we should either move the
Consumer to a modern version of Python (we know that part of the
library will not need to run on old Ubuntu versions as machine
charms) or pin the whole thing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,5 @@ git+https://github.com/canonical/operator/#egg=ops
 pyyaml
 urllib3
 jinja2 < 3
+markupsafe == 2.0.1
 kubernetes


sed-i (approved):
BTW, I suspect the itest will pass if you also pin pytest-operator:
`pytest-operator==0.10.1`

benhoyt:
Would be great to get this merged as it's breaking downstream CI as well. For example, I've hacked around it in https://github.com/canonical/operator/pull/668/files#diff-d899f68301edb06393d5435313801f676fc892f1484aa76f10d82c36e5e022f2R26

dstathis (approved):

PR #69, opened 2022-02-21
Pin version of markupsafe to fix build error

benhoyt:
This fixes the error when running the unit tests:

```
ImportError: cannot import name 'soft_unicode' from 'markupsafe'
```

This is needed due to a breaking change in the markupsafe module. See https://github.com/pallets/markupsafe/issues/284.

I'm not entirely sure it's the *right* fix. It would probably be better to pin all our dependency versions, but this seems like a good start.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 git+https://github.com/canonical/operator/#egg=ops
 pyyaml
 urllib3
-jinja2
+jinja2 < 3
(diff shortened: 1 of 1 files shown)


benhoyt:
Oops, sorry I should have checked open PRs! This has already been fixed (in the same way) in https://github.com/canonical/grafana-k8s-operator/pull/68 -- closing this.

PR #70, opened 2022-02-24
Don't use git for python-libjuju

rbarry82:


sed-i (approved):

dstathis (approved):
Consider changing pytest-operator as well

PR #73, opened 2022-03-03
Don't squash specially named datasources

rbarry82:
If dashboard templates already have `prometheusds` or `lokids`,
leave them alone

### DIFF
--- a/tests/unit/test_dashboard_consumer.py
+++ b/tests/unit/test_dashboard_consumer.py
@@ -403,3 +403,23 @@ class TestDashboardConsumer(unittest.TestCase):
                 }
             ],
         )
+
+    def test_consumer_templates_dashboard_and_keeps_datasources(self):
+        self.assertEqual(len(self.harness.charm.grafana_consumer._stored.dashboards), 0)
+        self.assertEqual(self.harness.charm._stored.dashboard_events, 0)
+        self.setup_different_dashboard(EXISTING_DATASOURCE_DASHBOARD_TEMPLATE)
+        self.assertEqual(self.harness.charm._stored.dashboard_events, 1)
+
+        # Comparing lists of dicts is painful. Convert back to a dict so we can sort
+        # and compare appropriately
+        db_content = json.loads(self.harness.charm.grafana_consumer.dashboards[0]["content"])
+        expected_content = json.loads(EXISTING_DATASOURCE_DASHBOARD_RENDERED)
+
+        db_content["templating"]["list"] = sorted(
+            db_content["templating"]["list"], key=lambda k: k["name"]
+        )
+        expected_content["templating"]["list"] = sorted(
+            expected_content["templating"]["list"], key=lambda k: k["name"]
+        )
+
+        self.assertEqual(db_content, expected_content)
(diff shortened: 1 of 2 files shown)


tests/unit/test_dashboard_consumer.py
```diff
@@ -126,6 +126,58 @@
     }
 )
 
+EXISTING_DATASOURCE_DASHBOARD_TEMPLATE = """
```
sed-i, line 129: These seem to be unused. Missing commit?

rbarry82: No, just not added yet -- I wanted you to be able to build to unblock.

Comparing the equality of a list of dicts in Python is absolutely terrible.

sed-i (approved):
Works. Hopefully merged soon so I won't have to change the load test's cloud-config.

mmanciop:
@rbarry82 I thought we had discussed the logic to be: “If there is precisely one Prometheus datasource, use that whatever its name is”, and to error if a dashboard has TWO datasources of type Prometheus. (And ofc the same logic would work with Loki.) The flexibility in the name of the datasource is very much needed for being able to reuse stock Grafana dashboards.

sed-i (dismissed):

rbarry82:
> @rbarry82 I thought we had discussed the logic to be: “If there is precisely one Prometheus datasource, use that whatever its name is”, and to error if a dashboard has TWO datasources of type Prometheus. (And ofc the same logic would work with Loki.) The flexibility in the name of the datasource is very much needed for being able to reuse stock Grafana dashboards.

The logic in the previous patch was actually "if a dashboard is imported and the name is _not_ `prometheusds|lokids` and it has the appropriate type, rename it so the dropdowns can find it". 

This was added explicitly for the case of "reuse stock Grafana dashboards" where the "template wizard" from the Grafana marketplace isn't used and the JSON is added directly, or in the case of `cos-config`, where it may be "there is a template variable defined for a Prometheus but it doesn't have the name we expect".

No change in behavior other than "don't show annoying error messages in the Grafana UI because your datasource was named `ds_prometheus` or whatever the default from the marketplace is"

mmanciop:
> if a dashboard is imported and the name is not prometheusds|lokids and it has the appropriate type, rename it so the dropdowns can find it

I feel like I am being dense at this stage, but hey I’ll go for broke: do you rename the datasource variable, or do you rename the reference to it in the variable dropdowns with the Juju topology?

rbarry82:
> > if a dashboard is imported and the name is not prometheusds|lokids and it has the appropriate type, rename it so the dropdowns can find it
> 
> I feel like I am being dense at this stage, but hey I’ll go for broke: do you rename the datasource variable, or do you rename the reference to it in the variable dropdowns with the Juju topology?

The reference in the dropdown is always `${prometheusds}` (which was chosen to avoid conflicting with `DS_PROMETHEUS`).

sed-i (dismissed):

mmanciop:
I think this PR should address #84

mmanciop:
I think this PR should address #84

rbarry82:
Closes #84

mmanciop:
> Closes #84

My man, shouldn't the offending dashboard be added to the tests? ;-)

Abuelodelanada (approved):

mmanciop (approved):

PR #76, opened 2022-03-15
Make sure Grafana is up before hitting the API to check password validity

rbarry82:
Since an HTTPError is also checked for in the password action, it's
possible to encounter a race where the action informs users that the
password has been changed when Grafana is simply not available yet.

No tests since Harness cannot check actions and we can't reliably reproduce a race in the integration tests.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -651,6 +651,7 @@ class GrafanaCharm(CharmBase):
         """Returns the password for the admin user as an action response."""
         if not self.grafana_service.is_ready:
             event.fail("Grafana is not reachable yet. Please try again in a few minutes")
+            return
         if self.grafana_service.password_has_been_changed(
             self.model.config["admin_user"], self._get_admin_password()
         ):
--- a/src/charm.py
+++ b/src/charm.py
@@ -649,6 +649,12 @@ class GrafanaCharm(CharmBase):
 
     def _on_get_admin_password(self, event: ActionEvent) -> None:
         """Returns the password for the admin user as an action response."""
+        if not self.grafana_service.is_ready:
+            event.set_results(
+                {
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -649,6 +649,12 @@ def _generate_datasource_config(self) -> str:
 
     def _on_get_admin_password(self, event: ActionEvent) -> None:
         """Returns the password for the admin user as an action response."""
+        if not self.grafana_service.is_ready:
+            event.set_results(
+                {
+                    "admin-password": "Grafana is not reachable yet. Please try again in a few minutes"
+                }
```
sed-i, line 656:
```suggestion
                {
                    "admin-password": "",  # or omit key entirely
                    "error": "Grafana is not reachable yet. Please try again in a few minutes"
                }
```

sed-i: Or even better: `event.fail()`

sed-i (changes requested):

sed-i (dismissed):

mmanciop (dismissed):
Change looks OK, but it deserves a unit test. It should be easy to do with a self.harness.begin() and directly invoking the action method from self.harness.charm passing a mock of ActionEvent

mmanciop:
> I should have clarified :| Of course, it's a Turing-complete language, anything _can_ be done. Whether or not we _should_ attempt to fake out actions by manually invoking private methods and mocking `ActionEvent` is another question.

Tbh this solution does not horrify me much. AFAICT OF does comparably little lifting for actions, as opposed to relations.

> I'll write the test here, but we should really make an effort to push on OF to implement actions in harness before this becomes a pattern.

Yep we should.

mmanciop (approved):
:heart:

sed-i (approved):

mmanciop (approved):

PR #80, opened 2022-03-16
Adds config to enable auth proxy

gruyaume:
Adds config to enable auth proxy
- Adds config for Grafana to let a HTTP reverse proxy handle authentication.
- Improves how Grafana's configuration file is generated based on Juju configs and relationship data

Explanation:
- Auth proxy is used in Magma where Grafana is shown within the application's UI frame and authentications is handled by passport.

Reference:
- Auth proxy authentication: https://grafana.com/docs/grafana/latest/auth/auth-proxy/

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -29,3 +29,10 @@ options:
       host or a bare A record, this may be omitted.
     type: string
     default: ""
+  enable_auth_proxy:
+    description: |
+      You can configure Grafana to let a HTTP reverse proxy handle authentication. Popular web 
+      servers have a very extensive list of pluggable authentication modules, and any of them can 
+      be used with the AuthProxy feature.
+    type: boolean
+    default: false
--- a/src/charm.py
+++ b/src/charm.py
@@ -358,6 +358,11 @@ class GrafanaCharm(CharmBase):
         rel = self.model.get_relation(DATABASE)
         return len(rel.units) > 0 if rel is not None else False
 
+    @property
+    def auth_proxy_enabled(self) -> bool:
+        """Returns whether or not enable_auth_proxy config is set."""
+        return self.model.config.get("enable_auth_proxy", False)
+
     def _on_database_changed(self, event: RelationChangedEvent) -> None:
         """Sets configuration information for database connection.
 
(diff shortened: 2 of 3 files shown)


mmanciop (changes requested):
I feel handling authentication through a reverse proxy belongs to the interplay with an Ingress charm. We are looking this sprint into integrating grafana-k8s with traefik-k8s, although we have not yet looked into the Auth aspect of it. I think we should not add this configuration option, but rather discuss the use case in the scope of ingress.

gruyaume:
> I feel handling authentication through a reverse proxy belongs to the interplay with an Ingress charm. We are looking this sprint into integrating grafana-k8s with traefik-k8s, although we have not yet looked into the Auth aspect of it. I think we should not add this configuration option, but rather discuss the use case in the scope of ingress.

I agree that making auth proxy configurable through a relationship makes sense. However, I don't think that it has to belong in the ingress. For example in our case we don't use ingress and need this auth proxy configured.

mmanciop:
> > I feel handling authentication through a reverse proxy belongs to the interplay with an Ingress charm. We are looking this sprint into integrating grafana-k8s with traefik-k8s, although we have not yet looked into the Auth aspect of it. I think we should not add this configuration option, but rather discuss the use case in the scope of ingress.
> 
> I agree that making auth proxy configurable through a relationship makes sense. However, I don't think that it has to belong in the ingress. For example in our case we don't use ingress and need this auth proxy configured.

I understand that you have this need, but exposing the configuration like this, for every Juju admin to wonder about, is not right either. You actually seem to use an ingress-like charm, namely https://github.com/canonical/charmed-magma/tree/main/orchestrator-bundle/nms-magmalte-operator . Since that is running a reverse proxy in front of Grafana (and other stuff as well), it could very well implement the provider side of the `ingress` relation.

gruyaume:
> > > I feel handling authentication through a reverse proxy belongs to the interplay with an Ingress charm. We are looking this sprint into integrating grafana-k8s with traefik-k8s, although we have not yet looked into the Auth aspect of it. I think we should not add this configuration option, but rather discuss the use case in the scope of ingress.
> > 
> > 
> > I agree that making auth proxy configurable through a relationship makes sense. However, I don't think that it has to belong in the ingress. For example in our case we don't use ingress and need this auth proxy configured.
> 
> I understand that you have this need, but exposing the configuration like this, for every Juju admin to wonder about, is not right either. You actually seem to use an ingress-like charm, namely https://github.com/canonical/charmed-magma/tree/main/orchestrator-bundle/nms-magmalte-operator . Since that is running a reverse proxy in front of Grafana (and other stuff as well), it could very well implement the provider side of the `ingress` relation.
> 
> Now, for completeness: if you did retrofit https://github.com/canonical/charmed-magma/tree/main/orchestrator-bundle/nms-magmalte-operator to implement the `ingress` relation, you still would not get the enablement of the auth proxy out of it. But that is a much better discussion to be had in the scope of "How do ingress-like charms relate with authentication of their downstreams?"

Yes I fully agree about not having this exposed through a config. Let me know how I can help with this ingress relationship you are referring.

For reference, here's the complete grafana.ini config that is used in our context:

```
[users]
allow_sign_up = false
auto_assign_org = false

[auth.proxy]
enabled = true
header_name = X-WEBAUTH-USER
header_property = username
auto_sign_up = false

[server]
root_url = /grafana
```

PR #82, opened 2022-03-21
cleanup automation and use current versions

simskij:
### Changes
- Replaces the legacy library check with the one provided by charming-actions.
- Replaces the legacy tagging mechanism with the one from charming-actions.
- Harmonizes the job names to match our other repositories.

### Test Instructions
- Verify that the workflow passes for the PR.
- Verify that the release on Charmhub is still valid after the release workflow has finished.

### DIFF
--- a/.github/workflows/release-edge.yaml
+++ b/.github/workflows/release-edge.yaml
@@ -1,4 +1,4 @@
-name: Release to edge
+name: Release to Edge
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-edge.yaml
```diff
@@ -1,109 +1,120 @@
...
-    - name: Run integration tests
-      run: tox -vve integration
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          juju-channel: latest/candidate # FIXME: Once 2.9.23 hits stable
```
sed-i, line 108: `2.9.26` is stable. Should be ok now.

sed-i (approved):

.github/workflows/release-edge.yaml
rbarry82: Realistically, we may want to have it as `FIXME: Once we go through 5 releases in a row without it breaking`

rbarry82 (approved):

mmanciop (approved):

PR #85, opened 2022-03-25
Add support for unit ingresses to GrafanaSourceProvider

rbarry82:
## Issue
Previously, `GrafanaSourceProvider` only supported connecting directly
to the root URL, so some source (let's say Prometheus) was no longer
accessible once ingress was enabled.

## Solution
A new argument was added to the constructor to allow specifying `source_uri`,
which will be used. If both `source_uri` and `source_port` are specified, a warning
will be emitted. If no scheme is set (such as `http://`), a warning will be emitted, and
the scheme added, so it is clear whether HTTP, HTTPS, or something else should
be used.

## Context
None.

## Testing Instructions
Enable ingress for Prometheus. Instantiate the Provider with the ingress path. Ensure
that it's usable from Grafana. 

Since the Provider Charm is aware of whether or not subpath routing needs to be used
if an ingress is set, even if it's an in-model connection, the Providing charm decides
whether or not to set this argument based on context. A PR is available for Prometheus

## Release Notes
Add support for unit ingresses/source paths  to GrafanaSourceProvider

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -365,9 +365,12 @@ class GrafanaSourceProvider(Object):
         self._relation_name = relation_name
         events = self._charm.on[relation_name]
 
-        refresh_event = refresh_event or self._charm.on.pebble_ready
-
         self._source_type = source_type
+
+        if not refresh_event:
+            if len(self._charm.meta.containers) == 1:
+                container = list(self._charm.meta.containers.values())[0]
+                refresh_event = self._charm.on[container.name.replace("-", "_")].pebble_ready
 
         if source_port and source_uri:
             logger.warning(
@@ -386,7 +389,8 @@ class GrafanaSourceProvider(Object):
         self._source_uri = source_uri
 
         self.framework.observe(events.relation_joined, self._set_sources)
-        self.framework.observe(refresh_event, self._set_unit_details)
+        if refresh_event:
+            self.framework.observe(refresh_event, self._set_unit_details)
 
     def _set_sources(self, event: RelationJoinedEvent):
         """Inform the consumer about the source configuration."""
@@ -429,7 +433,6 @@ class GrafanaSourceProvider(Object):
                 uri = self._source_uri
             else:
                 address = self._charm.model.get_binding(relation).network.bind_address
-                print("ELSE")
                 if address:
                     uri = "{}:{}".format(str(address), self._source_port)
 
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -25,17 +25,16 @@ provider in a charm which `provides: grafana-source`, creation of a
 The default arguments are:
 
     `charm`: `self` from the charm instantiating this library
+    `source_type`: None
+    `source_port`: None
+    `source_url`: None
     `relation_name`: grafana-source
     `refresh_event`: A `PebbleReady` event from `charm`, used to refresh
         the IP address sent to Grafana on a charm lifecycle event or
         pod restart
-    `source_type`: ""
-    `source_port`: ""
-    `source_uri`: ""
 
-`source_uri` should be a fully-resolvable URI for a valid Grafana source.
-As only HTTP sources are supported via libraries in the current charm, this
-is likely to be `http://example.com/api` or similar.
+The value of `source_url` should be a fully-resolvable URL for a valid Grafana
+source, e.g., `http://example.com/api` or similar.
 
 If your configuration requires any changes from these defaults, they
 may be set from the class constructor. It may be instantiated as
@@ -47,7 +46,9 @@ follows:
         def __init__(self, *args):
             super().__init__(*args, **kwargs)
             ...
-            self.grafana_source_provider = GrafanaSourceProvider(self)
+            self.grafana_source_provider = GrafanaSourceProvider(
+                self, source_type="prometheus", source_port="9090"
+            )
             ...
 
 The first argument (`self`) should be a reference to the parent (datasource)
@@ -317,11 +318,11 @@ class GrafanaSourceProvider(Object):
     def __init__(
         self,
         charm: CharmBase,
+        source_type: str,
+        source_port: Optional[str] = "",
+        source_url: Optional[str] = "",
         refresh_event: Optional[BoundEvent] = None,
         relation_name: str = DEFAULT_RELATION_NAME,
-        source_type: Optional[str] = "",
-        source_port: Optional[str] = "",
-        source_uri: Optional[str] = "",
     ) -> None:
         """Construct a Grafana charm client.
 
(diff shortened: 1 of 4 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -31,6 +31,7 @@
         pod restart
     `source_type`: prometheus
     `source_port`: 9090
+    `source_uri`: ""
```
mmanciop, line 34: This seems to need more documentation here to explain for which use-cases it is for.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -361,14 +368,29 @@ def __init__(
...
 
         self._source_type = source_type
+
+        if source_port and source_uri:
+            logger.warning(
+                "Both `source_port` and `source_uri` were specified! Using "
+                "`source_uri` as the address."
+            )
+
+        if source_uri and not re.match(r"^\w+://", source_uri):
```
mmanciop, line 378: I like!

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -348,6 +352,9 @@ def __init__(
                 the DataSource type from the Grafana perspective.
             source_port: an optional (default `9090`) source port
                 required for Grafana configuration.
+            source_uri: an optional source uri which can be used if ingress for
```
mmanciop, line 355: It should state that, if not falsy, it overrides the `source_port` argument.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -402,12 +424,19 @@ def _set_unit_ip(self, _: Union[BoundEvent, RelationEvent]):
...
-                relation.data[self._charm.unit]["grafana_source_host"] = "{}:{}".format(
-                    str(address),
-                    self._source_port,
-                )
+            uri = None
+            if self._source_uri:
+                uri = self._source_uri
+            else:
+                address = self._charm.model.get_binding(relation).network.bind_address
+                print("ELSE")
```
mmanciop, line 432: This needs to go :-)

mmanciop (changes requested):
Overall looks good, very minor changes seem needed.

mmanciop:
I hope later today I can try integrate it in `prometheus-k8s`

mmanciop:
I think the `refresh_event` pattern really did not age well :-(

For Prometheus, we have multiple events when to trigger: When Pebble comes up, and when the ingress changes. So, now I am creating a custom event source for Prometheus just to tell the Grafana source when to store the data in the relation and that is not great. IMO we should just always put the data passed in the constructor in the relation.

lib/charms/grafana_k8s/v0/grafana_source.py
mmanciop: +1

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -314,6 +317,7 @@ def __init__(
         relation_name: str = DEFAULT_RELATION_NAME,
         source_type: Optional[str] = "prometheus",
         source_port: Optional[str] = "9090",
+        source_uri: Optional[str] = "",
```
mmanciop, line 320: Let's get rid of these defaults, they work for just ONE charm :-)

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: That's actually why [we warn if there's no scheme and add a default](https://github.com/canonical/grafana-k8s-operator/pull/85/files#diff-7ed482b6184969c9603036d352837ddec840d76473e57a583c8b3c5bc49f4cbeR382), but point taken.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -361,14 +368,29 @@ def __init__(
         refresh_event = refresh_event or self._charm.on.pebble_ready
 
         self._source_type = source_type
+
+        if source_port and source_uri:
+            logger.warning(
+                "Both `source_port` and `source_uri` were specified! Using "
+                "`source_uri` as the address."
```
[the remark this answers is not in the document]
rbarry82: No, it should not be. This is setting attributes in the constructor, which _is_ the action taken, as those are looked up later.

While it would be equally possible to move this message to that function, it's private, would get logged every time `refresh_event` fired, and it would not be obvious from the log where/why to resolve it.

tests/unit/test_source_consumer.py
```diff
@@ -152,6 +152,56 @@ def test_consumer_notifies_on_new_sources(self):
...
+            "source_type": "prometheus",
+            "url": "http://1.2.3.4:9090/some/path",
+            "unit": "prometheus/0",
+        }
+
+        sources = self.harness.charm.source_by_rel_id(rel_id)[0]
+
+        self.assertIsNotNone(sources)
+        self.assertEqual(dict(sources), completed_data)
+        self.assertEqual(self.harness.charm._stored.source_events, 2)
```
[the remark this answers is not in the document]
rbarry82: Pretty much everywhere, in every test, yes, because harness is not always idempotent.

tests/unit/test_source_consumer.py
```diff
@@ -152,6 +152,56 @@ def test_consumer_notifies_on_new_sources(self):
         self.assertEqual(dict(sources), completed_data)
         self.assertEqual(self.harness.charm._stored.source_events, 2)
 
+    def test_consumer_notifies_on_new_sources_with_url_without_path(self):
```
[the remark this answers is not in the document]
rbarry82: There are no subtests anywhere in the Grafana charm, and they're intentionally avoided until Harness is feature-complete.

mmanciop (dismissed):

mmanciop (approved):
LGTM now. I took the risk of removing the non-sensical default values related with Prometheus, they seem like huge footguns that we should not have.

mmanciop:
Waiting for the itest to go green, when I'll squash&merge, and ship the lib

mmanciop (changes requested):
Bug to troubleshoot, see https://github.com/canonical/grafana-k8s-operator/pull/85#issuecomment-1079661302

mmanciop (approved):
Evidence seems pretty conclusive that the bug lurks somewhere in `prometheus-k8s` or the `ingress` library of `trafik-k8s`.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -310,10 +318,11 @@ class GrafanaSourceProvider(Object):
     def __init__(
         self,
         charm: CharmBase,
+        source_type: str,
```
mmanciop, line 321: The change of order broke the itests, hence bab99df1328d145e091bd4de31a11c9484052f12.

But given that the only charms that use the `GrafanaSourceProvider` are ours, so be it.

PR #100, opened 2022-05-27
Replace bind_address with fqdn

sed-i:
## Issue
`bind_address` sometimes returns None. Recently observed during load test.

Crossref: [OPENG-785](https://warthogs.atlassian.net/browse/OPENG-785)

## Solution
Use `socket.getfqdn` instead of `bind_address`.

## Context
NTA

## Testing Instructions
Run integration tests.

## Release Notes
Replace bind_address with fqdn.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -128,6 +128,7 @@ events may not be needed.
 import json
 import logging
 import re
+import socket
 from typing import Any, Dict, List, Optional, Union
 
 from ops.charm import (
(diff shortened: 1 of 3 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -842,11 +842,12 @@ def _update_all_dashboards_from_dir(self, _: Optional[HookEvent] = None) -> None
...
 
             # Path.glob uses fnmatch on the backend, which is pretty limited, so use a
-            # lambda for the filter
-            for path in filter(
-                lambda p: p.is_file and p.name.endswith((".json", ".json.tmpl", ".tmpl")),
-                Path(self._dashboards_path).glob("*"),
-            ):
+            # custom function for the filter
+            def _is_dashbaord(p: Path) -> bool:
+                return p.is_file and p.name.endswith((".json", ".json.tmpl", ".tmpl"))
```
Abuelodelanada, line 847: If I remember well we said that we should avoid inner functions.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1510,10 +1511,12 @@ def _maybe_get_builtin_dashboards(self, event: RelationEvent) -> Dict:
             )
 
         if dashboards_path:
-            for path in filter(
-                lambda p: p.is_file and p.name.endswith((".json", ".json.tmpl", ".tmpl")),
-                Path(dashboards_path).glob("*"),
-            ):
+
+            def _is_dashbaord(p: Path) -> bool:
+                return p.is_file and p.name.endswith((".json", ".json.tmpl", ".tmpl"))
```
Abuelodelanada, line 1516: If I remember well we said that we should avoid inner functions.

Besides we are duplicating this filter function

Abuelodelanada (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: Yes and yes:
1. It started with mypy complaining about untyped lambda args (but it's impossible to type lambda args).
2. @rbarry82 didn't like my module-level `_is_dashboard`, so that's where we're at.

PR #105, opened 2022-07-08
Template out __input variables also

### LINKED ISSUE #79: Adapt to the variable names existing in the dashboard
### Bug Description

Today, we expect the Grafana dashboards to be uploaded to Grafana via the `grafana_dashboard` library to use `prometheusds`  and `lokids` as names for the datasources of type `prometheus` and `loki` respectively.

This is not an assumption that works for many dashboards on the [Grafana Dashboards](https://grafana.com/grafana/dashboards/) website and it introduces unnecessary friction for the charm author and Juju admins wanting to deploy stock Grafana dashboards through the [COS Configuration](https://charmhub.io/cos-configuration-k8s) charm.

We should revisit our logic to do the following when importing a dashboard:

1. Look up the datasource variables in `.__inputs` (using `jq` notation)
2. Rename all entries with `type == 'datasource'` and `pluginId == 'prometheus'` to `DS_PROMETHEUS`
3. Rename all entries with `type == 'datasource'` and `pluginId == 'loki'` to `DS_LOKI`
4. Remove duplicates in the datasources [1]
5. Update all widgets to use the new datasource names :-)

[1] it might be that for some unspecified reason the user might have linked multiple Prometheus datasources in a dashboard but that is not a use-case we should support.

### To Reproduce

1. Try to add this [Traefik 2 dashboard](https://grafana.com/grafana/dashboards/11462) to the [traefik-k8s-operator](https://github.com/canonical/traefik-k8s-operator) charm
2. Despair

### Environment

This plane of reality

### Relevant log output

```shell
Michele wails in MatterMost about it a lot
```

### Additional context

_No response_

rbarry82:
## Issue
Previously, `grafana_dashboard` did not mangle/translate variables from `__input`, which is primarily used to prompt for user interaction if a dashboard is imported manually. The expected use case was that charm authors would adjust the dashboard to meet needs, then add it to the charm. This is now resolved.

Closes #79 

## Solution
Add parsing for `__input` to dashboards for `GrafanaDashboardConsumer` _only_ (this will not change the behavior of `GrafanaDashboardAggregator`, where we don't expect new dashboards)

## Testing Instructions
Add a dashboard from the marketplace (such as Loki metrics) and make sure it works

## Release Notes
Template out __input variables also

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -541,17 +541,26 @@ def _convert_dashboard_fields(content: str) -> str:
     datasources = {}
     existing_templates = False
 
+    # If the dashboard has __inputs, get the names to replace them. These are stripped
+    # from reactive dashboards in GrafanaDashboardAggregator, but charm authors in
+    # newer charms may import them directly from the marketplace
+    if "__inputs" in dict_content:
+        for field in dict_content["__inputs"]:
+            if "type" in field and field["type"] == "datasource":
+                datasources[field["name"]] = field["pluginName"].lower()
+        del dict_content["__inputs"]
+
     # If no existing template variables exist, just insert our own
     if "templating" not in dict_content:
         dict_content["templating"] = {"list": [d for d in TEMPLATE_DROPDOWNS]}
     else:
         # Otherwise, set a flag so we can go back later
         existing_templates = True
-        for maybe in dict_content["templating"]["list"]:
+        for template_value in dict_content["templating"]["list"]:
             # Build a list of `datasource_name`: `datasource_type` mappings
             # The "query" field is actually "prometheus", "loki", "influxdb", etc
-            if "type" in maybe and maybe["type"] == "datasource":
-                datasources[maybe["name"]] = maybe["query"].lower()
+            if "type" in template_value and template_value["type"] == "datasource":
+                datasources[template_value["name"]] = template_value["query"].lower()
 
         # Put our own variables in the template
         for d in TEMPLATE_DROPDOWNS:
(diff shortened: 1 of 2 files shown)


dstathis (approved):

balbirthomas (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -541,17 +541,26 @@ def _convert_dashboard_fields(content: str) -> str:
     datasources = {}
     existing_templates = False
 
+    # If the dashboard has __inputs, get the names to replace them. These are stripped
+    # from reactive dashboards in GrafanaDashboardAggregator, but charm authors in
+    # newer charms may import them directly from the marketplace
+    if "__inputs" in dict_content:
+        for field in dict_content["__inputs"]:
+            if "type" in field and field["type"] == "datasource":
+                datasources[field["name"]] = field["pluginName"].lower()
```
sed-i, line 550: - `pluginName` is guaranteed to exist for `field["type"] == "datasource"`?
- add comment about `.lower()`? Should it be `.casefold` just in case?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -541,17 +541,26 @@ def _convert_dashboard_fields(content: str) -> str:
...
         # Otherwise, set a flag so we can go back later
         existing_templates = True
-        for maybe in dict_content["templating"]["list"]:
+        for template_value in dict_content["templating"]["list"]:
             # Build a list of `datasource_name`: `datasource_type` mappings
             # The "query" field is actually "prometheus", "loki", "influxdb", etc
-            if "type" in maybe and maybe["type"] == "datasource":
-                datasources[maybe["name"]] = maybe["query"].lower()
+            if "type" in template_value and template_value["type"] == "datasource":
+                datasources[template_value["name"]] = template_value["query"].lower()
```
sed-i, line 563: `casefold` instead of `lower`?

rbarry82: Are we expecting sset or other UTF-8 characters which need to be converted?

PR #106, opened 2022-07-14
Dev authproxy config

saltiyazan:
## Issue
To deploy Grafana as part of the Magma bundle some configuration needs to be changed since currently grafana-k8s doesn't work with Magma out of the box. The aim is to integrate Grafana with Magma and deploy Grafana as part of the charmed-magma bundle.
Configure Grafana to have `enable_auth_proxy = True` and `enable_auto_assign_org = False`

## Solution
Adds `enable_auth_proxy` and `enable_auto_assign_org` as options to the `config.yaml` file of the charm, and sets the default values `enable_auth_proxy = False` and `enable_auto_assign_org = True`. Overrides Grafana configuration with environment variables if the given values are different than the default ones.

## Testing Instructions

Deploy magma-orc8r while replacing the grafna-k8s charm in the bundle with the one that includes these changes with the necessary options:
- `web_external_url: "/grafana"`
- `enable_auth_proxy: true`
- `enable_auto_assign_org: false`

## Release Notes
Add `enable_auth_proxy` and `enable_auto_assign_org` options to configure Grafana


rbarry82 (approved):

rbarry82:
@saltiyazan can you please rebase on `main` or `merge` on main? There's a fix for many of the integration test failures

simskij:
We are not going to merge this PR as-is. The decision we've made is to instead support a generic grafana-auth interface, which in the future will allow us to also add support for other authentication methods than auth proxy, like oauth, saml, jwt etc.

PR #108, opened 2022-07-18
Use asyncio_mode='auto' or async fixtures don't actually run

rbarry82:
## Issue
Fix integration tests not awaiting fixtures, and instead trying to deploy a `coroutine`

Unpin `pytest-operator` while we're at it.


balbirthomas (approved):
Tested

Abuelodelanada (approved):

PR #110, opened 2022-08-03
enable resource limits

dstathis:
## Issue
Add resource limits

## Solution
Use the new library

## Testing Instructions
Check limits work

## Additional Context
https://github.com/canonical/observability-libs/pull/19

## Release Notes
Added resource limits to grafana-k8s as a config option

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -35,7 +35,13 @@ from charms.grafana_k8s.v0.grafana_source import (
     GrafanaSourceEvents,
     SourceFieldsMissingError,
 )
+from charms.observability_libs.v0.kubernetes_compute_resources_patch import (
+    K8sResourcePatchFailedEvent,
+    KubernetesComputeResourcesPatch,
+    adjust_resource_requirements,
+)
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
+from lightkube.models.core_v1 import ResourceRequirements
 from ops.charm import (
     ActionEvent,
     CharmBase,
@@ -46,7 +52,7 @@ from ops.charm import (
 )
 from ops.framework import StoredState
 from ops.main import main
-from ops.model import ActiveStatus, MaintenanceStatus
+from ops.model import ActiveStatus, MaintenanceStatus, BlockedStatus
 from ops.pebble import ConnectionError, Layer, PathError, ProtocolError
 
 from grafana_server import Grafana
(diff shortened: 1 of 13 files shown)


src/charm.py
```diff
@@ -35,7 +35,13 @@
     GrafanaSourceEvents,
     SourceFieldsMissingError,
 )
+from charms.observability_libs.v0.kubernetes_compute_resources_patch import (
+    K8sResourcePatchFailedEvent,
+    KubernetesComputeResourcesPatch,
+    adjust_resource_requirements,
+)
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
+from lightkube.models.core_v1 import ResourceRequirements
```
sed-i, line 44: This can be imported from the lib.

dstathis: I have changed it.

tests/integration/test_resource_limits.py
```diff
@@ -0,0 +1,50 @@
...
+                "get",
+                "pod",
+                "-o",
+                "json",
+                f"{app_name}-0",
+            ],
+            text=True,
+        )
+    )
+    container = list(filter(lambda x: x["name"] == "grafana", pod["spec"]["containers"]))
```
sed-i, line 48: I wrote [`get_podspec`](https://github.com/canonical/prometheus-k8s-operator/blob/102bfa4820eb3832923c7b2654e9e269bfa3a3ae/tests/integration/helpers.py#L232) for this, might save a few lines in the test for the price of some other lines in `helpers` :)

src/charm.py
```diff
@@ -698,6 +710,14 @@ def _generate_password(self) -> str:
         chars = string.ascii_letters + string.digits
         return "".join(secrets.choice(chars) for _ in range(12))
 
+    def _resource_reqs_from_config(self) -> ResourceRequirements:
+        limits = {"cpu": self.model.config.get("cpu"), "memory": self.model.config.get("memory")}
+        requests = {"cpu": "0.25", "memory": "200Mi"}
+        return adjust_resource_requirements(limits, requests, adhere_to_requests=True)
+
+    def _on_resource_patch_failed(self, event: K8sResourcePatchFailedEvent):
+        self.unit.status = BlockedStatus(event.message)
```
sed-i, line 719: - Hopefully there wouldn't be unintended interactions with other grafana status sets?
- In other charms I had the following under `_configure`.

```python
        if not self.resources_patch.is_ready():
            if isinstance(self.unit.status, ActiveStatus) or self.unit.status.message == "":
                self.unit.status = WaitingStatus("Waiting for resource limit patch to apply")
            return
```

sed-i (approved):

tests/integration/test_resource_limits.py
dstathis: It seems a bit much to pull in lightkube just for this.

src/charm.py
dstathis: Grafana already sets a maintenance status in "on_stop" which I think should be good enough here.

By the way, MaintenanceStatus is appropriate here. WaitingStatus is specifically for waiting on a relation.

sed-i: Seeing a record of limits update in `juju show-status-log` is valuable, imo.

Agreed, it should be Maintenance.

dstathis: Okay, I will add this.

rbarry82: Grafana doesn't tweak status often, so there's no negative interactions there.

sed-i (approved):

Abuelodelanada (approved):

PR #111, opened 2022-08-06
Fix password integration test

rbarry82:
## Issue
Make the integration tests for password validity work again.

Also update the libs, and unpin `python-libjuju` from integration tests

## Solution
After trying to bisect Juju to discover where this issue appeared, I wasn't able to find a root cause.

Instead, the very rapid stopping from the Juju/pebble bug [here](https://bugs.launchpad.net/juju/+bug/1951415) killed the containers so quickly on scaling (despite `force` being a null argument from `python-libjuju` to the facade to scale) that k8s was unhappy:

```
Events:                                                                                                                                                                                                                                         
  Type     Reason     Age               From               Message                                                                                                                                                                              
  ----     ------     ----              ----               -------
  Normal   Scheduled  64s               default-scheduler  Successfully assigned test-get-password-9dhe/grafana-0 to jujubisect
  Normal   Pulled     63s               kubelet            Container image "jujusolutions/jujud-operator:2.9.32" already present on machine
  Normal   Created    63s               kubelet            Created container charm-init
  Normal   Started    63s               kubelet            Started container charm-init
  Normal   Started    62s               kubelet            Started container charm
  Normal   Created    62s               kubelet            Created container charm
  Normal   Pulled     62s               kubelet            Container image "jujusolutions/charm-base:ubuntu-20.04" already present on machine
  Normal   Pulled     62s               kubelet            Container image "docker.io/ubuntu/grafana:8.2-22.04_beta" already present on machine
  Normal   Created    62s               kubelet            Created container grafana
  Normal   Started    62s               kubelet            Started container grafana
  Normal   Killing    24s               kubelet            Stopping container charm
  Normal   Killing    24s               kubelet            Stopping container grafana
  Warning  Unhealthy  4s (x5 over 24s)  kubelet            Readiness probe failed: Get "http://10.1.155.62:38813/v1/health?level=ready": dial tcp 10.1.155.62:38813: connect: connection refused
  ```
  
The unit was stuck in `1/0` forever, since the finalizer would not run. 

This was surfaced by the fact that `python-libjuju.model.wait_for_idle(...)` (in what certainly looks like a bug) simply does nothing if `wait_for_exact_units=0`, because none of the conditionals are hit. It's `> 0`, Juju still thinks the unit is active with an idle model due to the finalizer bug, and `wait_for_units` cannot trigger either unless we can trick `len(app.units)` into being negative (we can't).

This appears to have succeeded before because the action was actually running _prior_ to it shutting down, since it never blocked on anything! 🤔 

## Release Notes
Update `prometheus_scrape` library dependencies, fix the password integration test.

### DIFF
--- a/tests/integration/test_get_password.py
+++ b/tests/integration/test_get_password.py
@@ -2,7 +2,7 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-
+import asyncio
 import logging
 from pathlib import Path
 
@@ -22,7 +22,7 @@ async def test_setup_env(ops_test):
 
 
 @pytest.mark.abort_on_fail
-async def test_password_returns_correct_value_after_scaling(ops_test, grafana_charm):
+async def test_password_returns_correct_value_after_upgrading(ops_test, grafana_charm):
     """Deploy from charmhub and then upgrade with the charm-under-test."""
     logger.info("deploying local charm")
     await ops_test.model.deploy(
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by simskij — APPROVED** [review]

**Comment by simskij** [comment]
Once checks are passing, I'm happy with having this merged.

**Review by dstathis — APPROVED** [review]

File: tests/integration/test_get_password.py [hunk 1]
```diff
@@ -22,7 +21,7 @@ async def test_setup_env(ops_test):
 
 
 @pytest.mark.abort_on_fail
-async def test_password_returns_correct_value_after_scaling(ops_test, grafana_charm):
+async def test_password_returns_correct_value_after_upgrading(ops_test, grafana_charm):
```
sed-i, line 24: `correct_value_after_upgrading` is a good test on its own, but why remove the `scale->0->1`?

rbarry82: As mentioned on the call and the PR summary `scale->0` is not enforceable.

For one, no conditional in [`wait_for_idle`](https://github.com/juju/python-libjuju/blob/34981c59eac8db89320bf0067bebd2547da12d45/juju/model.py#L2476) would actually match this, and the fact that this ever worked before was simply a misleading test (which presumably ran the action before the charm stopped, since it wouldn't have blocked at all). 

We could build our own with `block_until`, but it's not meaningful since the k8s healthcheck fails due to the bug linked in the description, so it never scales down in the first place. It's just `1/0` until the model dies.

So there are two issues blocking a scale->0->anything test anyway, and I haven't tested, but the "pods get instantly killed with a failing healthcheck so it can't finalize" problem may actually fail any tests which scale down at all in any scenario.

sed-i: > `scale->0` is not enforceable.
> A `scale->0->1` test would never succeed until these issues are resolved

The scale to zero test came about due to a real bug, so we'd be losing coverage, iiuc.

> this test essentially exercises the same functionality.

I'm not clear on the relevant differences between upgrade and scale 0 -> 1. If there aren't any then +1 for merging. Otherwise perhaps keep the upgrade test and xfail the scale down-up?

rbarry82: > The scale to zero test came about due to a real bug, so we'd be losing coverage, iiuc.

We would not, which was the end result of trying to bisect when this stopped working. Due to the fact that none of conditionals in `wait_for_idle` make it possible to actually wait for zero units, this never actually worked the way it was intended, which is also true for any other tests which scale to zero.

PR #114, opened 2022-08-29
Dashboards links stable

Abuelodelanada:
## Issue

If a charm provides a dashboard with an `uid` Grafana will use that uid to create the URL, for instance a dashboard file with:

```json
...
  "title": "Zinc",
  "uid": "z9ghSzQ7z",
  "version": 1
}
```

Grafana will generate a stable URL like: `http://10.1.157.73:3000/d/z9ghSzQ7z/zinc?orgId=1`

The problem appears if the charm provides a dashboad **without** `uid`. 
In this situation Grafana will generate a `uid` based on the JSON file contents... if the JSON file does not change, every time the charm is packed,  refreshed (or deployed and related to Grafana), the URL will be the same.

But if the JSON file is modified (because for example the charm author added a new panel to the dashboard), the URL generated by Grafana will be different.

## Solution

A new private method in `GrafanaDashboardConsumer` was added. This method `_manage_dashboard_uid` will generate an `uid` if the JSON content hasn't.

## Testing Instructions

Run unit tests.

Manually:

- Deploy Grafana
- In [Zinc](https://github.com/jnsgruk/zinc-k8s-operator) charm delete `uid` in Grafana dashboad.
- Pack and deploy the charm
- Check dashboard URL generated by Grafana
- Modify the dashboard file.
- Re-pack and refresh Zinc charm
- Verify the Dashboard URL is the same.


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1221,6 +1222,19 @@ def _render_dashboards_and_signal_changed(self, relation: Relation) -> bool:  #
...
+    def _manage_dashboard_uid(self, dashboard: str, relation: Relation, fname: str) -> str:
+        """Add an uid to the dashboard if it is not present."""
+        dashboard = json.loads(dashboard)
+
+        if not dashboard.get("uid", None):
+            dashboards = json.loads(relation.data[relation.app]["dashboards"])  # type: ignore
+            dashboard["uid"] = "{}-{}".format(
+                dashboards["templates"][fname]["juju_topology"]["model_uuid"][:8],
+                relation.app.name[:6],  # type: ignore
+            )
```
sed-i, line 1234: Just wondering:
- would dashboard["uid"] ever by false-y is topology is present?
- for dashboards coming via cos-config, I do not expect to have any topology, which would result in a KeyError?

sed-i (commented):
- What do we mean by "stable" links? Something the SRE can bookmark once and for all?
- If this is only needed for keeping other relation data up to date, then if we're already ok with having a `uid` in the URL, can we use grafana's own algorithm for generating the uid from the json ourselves?
- Is there a way to avoid having the uid as part of the URL? Having the `uid` as part of the URL makes the URLs less "predictable".

lib/charms/grafana_k8s/v0/grafana_dashboard.py
simskij: Since the user will be able to pick the model from the drop downs, it feels like we should use something else as the uuid. How about the charm name? If there are multiple deployments, they'd end up sharing dashboards, which I think is fine. If there are multiple dashboards with the same name in different charms they'd stay differentiated.

Abuelodelanada:
> * What do we mean by "stable" links? Something the SRE can bookmark once and for all?

Yes, links that does not change over the time

 
>     * If this is only needed for keeping other relation data up to date, then if we're already ok with having a `uid` in the URL, can we use grafana's own algorithm for generating the uid from the json ourselves?

The uid generated by Grafana depends on the JSON content... so if the JSON file changes, the generated uid will change as well.


 
>     * Is there a way to avoid having the uid as part of the URL? Having the `uid` as part of the URL makes the URLs less "predictable".

As far as I know there is no way to avoid having the uid as parte of the URL.

For instance an `uid` generated with the Juju Topology will looks like:

`http://10.1.157.113:3000/d/61ceec42-cos-co/zinc?orgId=1`

lib/charms/grafana_k8s/v0/grafana_dashboard.py
Abuelodelanada: > Since the user will be able to pick the model from the drop downs, it feels like we should use something else as the uuid. How about the charm name? If there are multiple deployments, they'd end up sharing dashboards, which I think is fine. If there are multiple dashboards with the same name in different charms they'd stay differentiated.
> 
> In the case of cos-config, it could be `{ORG}-{REPO}-{BRANCH}` or similar, to keep it unique enough(ish)?

The problem we have here, is that the object that builds the `uid` is `GrafanaDashboardConsumer` that runs on Grafana side, and `ORG`, `REPO` and `BRANCH` are not shared over relation data by cos config.

From my point of view, in this situation, a "good" `uid` could be created out of:

`{model_id}-{app.name}-{file_name}`

`model_id` is unique. Inside a juju model, application name is unique as well, that's why I believe `charm_name` is not going to add value.

In the cos-config situation, `model_id` is the `id` of the model in which cos-config is running. File names will be unique as well.

rbarry82: The object that builds the `uid` actually isn't even `GrafanaDashboardConsumer`, to be really clear. Unless the dashboard JSON has a `uid` in it, the Grafana application assigns one itself.

The `uid` must be a string between 8 and 40 characters. Depending on how long the filename or git `{org}-{repo}-{branch}` may be, that may be too long. In general, my take (as on the call) is basically that any kind of hashing (sha, etc) of the template is not necessarily meaningful, because it may or may not have topology injected into it, and will definitely have a datasource picker injected into it, so it will change on the consumer side regardless.

That logic could theoretically move into the `Provider` side, but it would be a change in the flow/API, and if so, it would also involve modifying `GrafanaDashboardAggregator` for `cos-proxy`, which would be ideal to avoid.

The "same" dashboard (same directory structure, whether after a repo change in `cos-config` or a charm refresh/upgrade) should get the "same" UID, which is complicated by potentially overrunning 40 characters for the UID, and that chance increases the more complexity we add.

`GrafanaDashboardProvider` does not iteratively/blindly scan relations to build data on every event, unlike other charms.

Instead, it has a peer relation, which contains:
```
{
  {rel_id}: [
    {
      "id": "{relation.name}:{relation.id}/{filename}",
      "original_id": "{filename}",
      "content": "{rendered_template}",
      "template": "{original_dashboard_template}",
      "valid": [True|False],
      "error": "{errors_if_any}"
   },
   {
     ...
   },
]
```

So, we need a semi-unique scheme which fits into a constrained keyspace. How `cos-config` handles this is up to `cos-config` as long as it fits inside those constraints. That is:

* One or more `cos-config` charms related to a Grafana should yield identical UIDs for identical files
* One or more `foo` charms related to Grafana should yield the same UID for the same dashboard. Nominally, I would suggest that we go further than this, and if the _same_ dashboard (by filename, with topology set) is also present with a different `{id}`, that we keep both, because otherwise we _will_ eventually get a request to have both an "old" and a "new" version of some charm with different dashboards/metrics available simultaneously

This means that, in practice, we should plan for a mechanism (which does not need to be implemented right now) to "talk back" to any given provider so it can appropriately set AlertManager drill-down links, and/or potentially set a status message.

Given this, I would suggest using not `app.name` (which could easily differ between `prometheus` and `prometheus-2`), but `charm.meta.name`. `{charm.meta.name}-{hashlib.shake_256(path.encode()).hexdigest(4)` could be a good proxy, but which hash we use doesn't really matter.

Note that, as a matter of course at the moment, `JujuTopology` is not used, and if `JujuTopology` is added, care must be taken to also update `cos-proxy` if we're doing any reconstruction, or to provide a fallback method to construct `JujuTopology` objects from the existing keys to continue providing compatibility with the aggregator.

With that in mind, I would probably update the existing `_juju_topology` method to add `charm.meta.name` as an additional key, then add the hash chosen to each dashboard ([this PR](https://github.com/canonical/grafana-k8s-operator/pull/116) could be a useful proxy in terms of the changes to relation data if you look at the unit tests).

Following this, on the Provider side, we can blindly concatenate them into a UID, and, if no UID is already present in the JSON (so we don't stomp on UIDs set in dashboards from `cos-proxy`), insert it. We don't _have to_ do it this way, but doing so leaves the door open in the future for different dashboards by adding another peer data key with an inverted datastructure, so an added dashboard could say:

* (Provider) Here's a dashboard and the UID I want
* (Consumer) I see your request, but I have another dashboard with the same `charm.meta.name` and hash. I generated this one instead
* (Provider) Ack, I'll set a status message and/or log a message reflecting that change, and update my internal state so alerts deep link to the right place

Again, that part does not need to happen right now, but it will be a product request sooner or later, and we may as well design around that inevitability. So:

* Put `charm.meta.name` into the relation data along with a hash (long enough to be unique, short enough to not be painful)
* For now, we use that blindly. If there's no UID in the dashboard already, use this one. If there is a UID, keep that
* For now, assume that the `Provider` is the "source of truth" for what the UID will be and the Consumer will accept it. That assumption will probably change in the future

Given those assumptions, the uid generation can be _mostly_ on the `Provider` side, and the `Consumer` only needs to check "is there a UID? No? Insert it", with no message interchange.

`cos-proxy` could use any other proxy instead of `charm.meta.name`, but a combination of `charm.meta.name` and a hashed complete filepath should be sufficient both for uniqueness across identical filenames in deeply nested repos, yet distinct enough that multiple `cos-config` relations won't end up with duplicate dashboards

simskij: > model_id is unique. Inside a juju model, application name is unique as well, that's why I believe charm_name is not going to add value.

That is exactly why we want to use the charm name. We don't want duplicates of the exact same dashboard, so we only want it to be unique enough to guarantee we're not throwing away a different dashboard that happens to have the same name. :)

rbarry82: Think of it this way. For one thing, the model right now doesn't make any assumptions about whether or not the dashboard from the `Provider` is actually a Jinja template, but...

More explicitly, validation of duplicate uid hashes can only be done from a place where there is full visibility of all dashboards, and that is the Consumer.

Given an arbitrary:
* model_a
  - foo_charm
    - dashboard.json -> `0xdeadbeef`
  - foo2_charm
    - dashboard.json -> `0xdeadbeef`
 * model_b
   - foo_charm
     - dashboard.json -> `0xdeadbeef`
   - old_foo_charm
     - dashboard.json -> `0xfeel1dead`
     
In all of these, only the `Consumer` has a view of all dashboards, and only the `Consumer` can confirm or deny "you requested this `{name}-{hash}` and the actual value is the same, so I'll accept it and let you go on your way" or "you requested this `{name}-{hash}` but I already have that present, and a more thorough hash of the file says that it's actually different, so use this one instead"

Generating it in the `Provider` means less cross-talk assuming a "positive" result (no collisions), but leaves an "escape hatch" in case there is a need later.

rbarry82 (approved):

tests/unit/test_dashboard_consumer.py
```diff
@@ -324,6 +324,15 @@ def test_consumer_notifies_on_new_dashboards(self):
...
 
+        # print(self.harness.charm.grafana_consumer.dashboards)
+        # print([
+        #         {
+        #             "id": "file:tester",
+        #             "relation_id": "2",
+        #             "charm": "grafana-k8s",
+        #             "content": DASHBOARD_RENDERED,
+        #         }
+        #     ])
```
sed-i, line 335: Is this comment intended here?

sed-i (approved):

tests/unit/test_dashboard_consumer.py
Abuelodelanada: Oh, I forget it!

Done!

PR #115, opened 2022-08-30
creates grafana-auth lib

saltiyazan:
## Issue
This PR creates a library called `grafana-auth` that implements [grafana-auth relation interface](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/grafana_auth/v0/README.md).
This library addresses the need to configure the authentication mode to authenticate with `Grafana`.

## Solution
The library allows the `provider` charm to pass authentication configuration that will be used by the `requirer` charm to authenticate with `Grafana`. 
Now it allows to setup the `Proxy` authentication mode.

## Testing Instructions
Unit and integration testing is added.


lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+        self.relation_id = snapshot["relation_id"]
+
+
+class AuthProviderCharmEvents(CharmEvents):
+    """List of events that the auth provider charm can leverage."""
+
+    urls_available = EventSource(UrlsAvailableEvent)
+
+
+class AuthProvider(Object):
```
Abuelodelanada, line 273: If I understand correctly, in the docstrings we named this class: `GrafanaAuthProxyProvider`

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+            "relation_id": self.relation_id,
+        }
+
+    def restore(self, snapshot: dict):
+        """Restores snapshot."""
+        self.urls = _type_convert_stored(snapshot["urls"])
+        self.relation_id = snapshot["relation_id"]
+
+
+class AuthProviderCharmEvents(CharmEvents):
```
Abuelodelanada, line 267: I would name this class `GrafanaAuthProviderCharmEvents`

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+
+    def __init__(self, charm: CharmBase, relationship_name: str):
+        super().__init__(charm, relationship_name)
+        self._auth_config = {}  # type: Dict[str, Dict[str, Any]]
+        self._charm = charm
+        self._relationship_name = relationship_name
+        container = list(self._charm.meta.containers.values())[0]
+        if len(self._charm.meta.containers) == 1:
+            refresh_event = self._charm.on[container.name.replace("-", "_")].pebble_ready
+            self.framework.observe(refresh_event, self._get_urls_from_relation_data)
```
Abuelodelanada, line 286: What happens if our Grafana charm has more than one container?

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+            self._charm.on.leader_elected, self._set_auth_config_in_relation_data
+        )
+        self.framework.observe(
+            self._charm.on[relationship_name].relation_changed, self._get_urls_from_relation_data
+        )
+
+    def _set_auth_config_in_relation_data(
+        self, event: Union[LeaderElectedEvent, RelationJoinedEvent]
+    ) -> None:
+        """Handler triggered on relation joined event.
```
Abuelodelanada, line 301: Not only on relation joined.

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+class AuthProvider(Object):
+    """Authentication configuration provider class to be initialized by auth providers."""
+
+    on = AuthProviderCharmEvents()
+
+    def __init__(self, charm: CharmBase, relationship_name: str):
+        super().__init__(charm, relationship_name)
+        self._auth_config = {}  # type: Dict[str, Dict[str, Any]]
+        self._charm = charm
+        self._relationship_name = relationship_name
```
Abuelodelanada, line 282: In all libs we are using `relation_name`

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+
+        relation = self._charm.model.get_relation(self._relationship_name)
+        if not relation:
+            logger.warning("Relation {} has not been created yet".format(self._relationship_name))
+            return
+
+        if not self._auth_config:
+            logger.warning(
+                "No authentication configuration was given by for application {} , it won't be set in relation {}".format(
+                    self.model.app, relation.id
```
Abuelodelanada, line 322: When passing messages to the logger, do not build the strings yourself. Take a look at: https://juju.is/docs/sdk/logging

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+class AuthProviderCharmEvents(CharmEvents):
+    """List of events that the auth provider charm can leverage."""
+
+    urls_available = EventSource(UrlsAvailableEvent)
+
+
+class AuthProvider(Object):
+    """Authentication configuration provider class to be initialized by auth providers."""
```
Abuelodelanada, line 274: Improve docstrings here.
Take a look at this example: https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L1447

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+        urls = json.loads(urls_json)
+
+        self.on.urls_available.emit(urls=urls, relation_id=relation.id)
+
+    def _validate_auth_config_json_schema(self) -> bool:
+        """Validates authentication configuration using json schemas.
+        Returns:
+            bool: Whether the configuration is valid or not based on the json schema.
+        """  # noqa
+        return False
```
Abuelodelanada, line 369: Is this method in a WIP state?

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+class AuthRequirerCharmEvents(CharmEvents):
+    """List of events that the auth requirer charm can leverage."""
+
+    auth_conf_available = EventSource(AuthConfAvailableEvent)
+
+
+class AuthRequirer(Object):
+    """Authentication configuration requirer class to be initialized by auth requirers."""
```
Abuelodelanada, line 400: What about naming this class as: `GrafanaAuthRequirer`?

Improve dosctrings. An example: https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L1447

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+
+    on = AuthRequirerCharmEvents()
+
+    def __init__(
+        self,
+        charm,
+        urls: List[str],
+        relationship_name: str = DEFAULT_RELATION_NAME,
```
Abuelodelanada, line 408: We use `relation_name` in our other libs

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+    ):
+        super().__init__(charm, relationship_name)
+        self._charm = charm
+        self._relationship_name = relationship_name
+        self._urls = urls
+        container = list(self._charm.meta.containers.values())[0]
+        if len(self._charm.meta.containers) == 1:
+            refresh_event = self._charm.on[container.name.replace("-", "_")].pebble_ready
+            self.framework.observe(refresh_event, self._get_auth_config_from_relation_data)
```
Abuelodelanada, line 417: What happens if our Consumer charm has more than one container?

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
...
+        Returns:
+            None
+        """  # noqa
+        super().__init__(charm, relationship_name)
+        self._auth_config[self._AUTH_TYPE] = {}
+        self._auth_config[self._AUTH_TYPE]["enabled"] = self._ENABLED
+        self._auth_config[self._AUTH_TYPE]["header_name"] = header_name
+        self._auth_config[self._AUTH_TYPE]["header_property"] = header_property
+        self._auth_config[self._AUTH_TYPE]["auto_sign_up"] = auto_sign_up
+        if sync_ttl is not None:
```
Abuelodelanada, line 550: `if sync_ttl is not None:` -> `if sync_ttl:`


The same pattern can be used in the next ifs

Abuelodelanada (dismissed):

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: I had `Grafana` as a prefix for most if not all of the class names and functions in the library. But based on some discussions, I removed the word `Grafana` wherever that is possible, to make this library more general.
I spotted some places in docstrings where `Grafana` was not yet removed so I fixed that now.
`GrafanaAuthProxyProvider` is `Grafana` specific and the library can be extended with similar classes.
However the requirer `AuthRequirer` seems to be independent from `Grafana` and the base provider `AuthProvider` also seems to be independent from `Grafana` but encapsulating the common logic that a provider of any authentication mode to any service can use.

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: As mentioned above, I tried to avoid the word `Grafana` where there is no dependency on `Grafana` to make the lib more general.

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: The base class `AuthProvider` shouldn't be initialised and this function of the base class shouldn't be called (it's implemented in the child class).
I changed it to raise `NotImplementedError` and modified the docstring maybe that provides a better feedback for the user of this library.
Please note that I'm avoiding using `abc` :)
let me know if you have different thoughts, I'm okay with changing this pattern if it's not pythonic enough :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: `AuthProvider` is the base class and `GrafanaAuthProxyProvider` is the child class which creates a configuration with `proxy authentication` mode for `Grafana`.
I modified the docstring to explain this :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: oh, good catch :)
Fixed following the idea in this [example](https://github.com/canonical/grafana-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: Fixed like above :)

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,629 @@
...
+            self.grafana_auth_proxy_provider.on.urls_available, self._on_urls_available
+        )
+        ...
+```
+Values different than defaults must be set from the class constructor.
+The [official documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/)
+of Grafana provides further explanation on the values that can be assigned to the different variables.
+Example:
+```python
+from charms.grafana_auth.v0.grafana_auth import GrafanaAuthProxyProvider
```
sed-i, line 54: Following up on a [comment elsewhere](https://github.com/canonical/grafana-auth-interface/pull/1#discussion_r948179621), I am still curious if having a thinner wrapper around each type of auth would be easier.

I.e. have one `GrafanaAuthProvider`, and [multiple types](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#authanonymous) of auth objects it can take in the constructor arg.

For example:
```python
self.auth = GrafanaAuthProvider(self, method=AuthProxy.from_dict(self.config))
# or
self.auth = GrafanaAuthProvider(self, method=AuthGithub.from_dict(self.config))
# etc.
```

tests/integration/auth-provider-tester/lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
+# Copyright 2022 Canonical Ltd.
```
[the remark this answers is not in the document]
saltiyazan: Fixed

tests/integration/grafana-tester/lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,573 @@
+# Copyright 2022 Canonical Ltd.
```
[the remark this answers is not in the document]
saltiyazan: Fixed

tests/integration/test_grafana_auth.py
```diff
@@ -0,0 +1,90 @@
...
+        await self._deploy_requirer(ops_test)
+
+    def _find_charm(self, charm_dir: str, charm_file_name: str) -> Optional[str]:
+        for root, _, files in os.walk(charm_dir):
+            for file in files:
+                if file == charm_file_name:
+                    return os.path.join(root, file)
+        return None
+
+    async def _deploy_provider(self, ops_test):
```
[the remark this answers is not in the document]
saltiyazan: Fixed

tests/unit/test_auth_proxy_provider.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @patch(
+        f"{CHARM_LIB_PATH}.AuthProvider._get_urls_from_relation_data",
+    )
+    def test_given_charm_has_one_container_and_refresh_event_not_specified_when_initialized_then_pebble_ready_is_set_as_refresh_event(
+        self,
+        mock_get_auth_config_from_relation_data,
+    ):
+        self.harness.container_pebble_ready("auth-tester")
+        mock_get_auth_config_from_relation_data.assert_called_once()
```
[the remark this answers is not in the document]
saltiyazan: Fixed

tests/unit/test_auth_proxy_provider.py
```diff
@@ -0,0 +1,233 @@
...
+    def setUp(self):
+        self.harness = testing.Harness(DefaultAuthProxyProviderCharm, meta=METADATA)
+        self.harness.set_leader(True)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    @patch(
+        f"{CHARM_LIB_PATH}.AuthProvider._get_urls_from_relation_data",
+    )
+    def test_given_charm_has_one_container_and_refresh_event_not_specified_when_initialized_then_pebble_ready_is_set_as_refresh_event(
```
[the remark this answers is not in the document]
saltiyazan: Fixed

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,629 @@
...
+# See LICENSE file for licensing details.
+
+"""grafana-auth library.
+
+This library implements the grafana-auth relation interface,
+it contains the Requirer and Provider classes for handling
+the interface.
+This library is designed to allow charms to configure authentication to Grafana,
+the provider will set the authentication mode that it needs,
+and will pass the necessary configuration of that authentication mode.
```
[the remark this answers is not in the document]
saltiyazan: Fixed

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,629 @@
...
+
+        Returns:
+            None
+        """
+        if not self._charm.unit.is_leader():
+            return
+
+        relation = self._charm.model.get_relation(self._relation_name)
+        if not relation:
+            logger.warning("Relation %s has not been created yet", self._relation_name)
```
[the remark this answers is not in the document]
saltiyazan: Yes, I agree. It it is actually better to give the charm the responsibility of assuring that the relation is created at some point. 
I removed it.

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,629 @@
...
+            return
+
+        if not self._auth_config:
+            logger.warning(
```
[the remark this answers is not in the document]
saltiyazan: Same as above, I agree that the charm should check if it gets the config, and block if it doesn't, and it could log something similar :). I removed it.

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,629 @@
...
+                "No authentication configuration was given by the application, it won't be set in relation %d",
+                relation.id
+                )
+            return
+        if not self._validate_auth_config_json_schema():
+            logger.warning(
```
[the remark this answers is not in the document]
saltiyazan: you are right, it isn't enough, `return` was missing :)
I added it, however I would keep the log messages here as the charm won't know this information to be able to log it on its side :)

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: If I understand the proposal correctly the parameter `method` would be of type `AuthType` or `AuthMethod` which would be the base class for the different authentication types like `AuthProxy` and `AuthGithub`. If that's the case then we haven't avoided inheritance, but we added an extra layer which is the `GrafanaAuthProvider` wrapper.
Could you please help me understand what I might be missing? As I don't see a clear advantage for this solution over the current approach.

sed-i: I mean avoid multilevel inheritance from `Object`.
Currently the `Object`-derived object is GrafanaAuthProvider, and it is tightly coupled to both operator framework (as expected) and the auth yaml config block construction. The latter logic doesn't have to be baked into the GrafanaAuthProvider constructor.

If possible, I think we could benefit here from a stronger separation of concerns: one class for constructing a yaml config block, and another class for managing the relation. If possible.

lib/charms/grafana_auth/v0/grafana_auth.py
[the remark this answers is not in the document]
sed-i: If this is important enough then could emit a custom event, which the charm would observe and block if necessary.

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: I get it now, thanks :)
Yeah in this case I agree, it does look cleaner.
However the current approach makes the usage of the library very easy, the user would only need to have something like `self.grafana_auth_config = GrafanaAuthProxyProvider(self)` if the default parameters are okay.
With the new approach the user would have to instantiate 2 objects. (still fairly easy tho).

Just to double check with Ryan as well, as he was involved in the discussion on this design.
What do you think @rbarry82 ?

lib/charms/grafana_auth/v0/grafana_auth.py
```diff
@@ -0,0 +1,614 @@
+# Copyright 2022 Canonical Ltd.
```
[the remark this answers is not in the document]
rbarry82: AFAIK, there's no convention against having "disconnected" libs, but I'm happy to wait on moving it into `charms/grafana_k8s` for now. If this becomes a "more generalized" auth library for o11y charms, keeping it isolated from the "main" grafana-k8s charmhub page also makes migrating it to its own repo easier than having an "old", deprecated library published.

lib/charms/grafana_auth/v0/grafana_auth.py
rbarry82: Multilevel inheritance from `Object` is fine, considering how `obj.handle` is constructed by Operator Framework, and this sort of multi-level inheritance is something we actually do inside `loki_push_api` to sync up comon functionality.

Personally, as mentioned earlier, the `self.library_instance = Library(self)` is a common pattern used all over the o11y libs which we intentionally designed and should stick to.

`GrafanaAuthProvider(self, method: dict...)` isn't any more fluent or user friendly than just accepting a raw dict, regardless of whether there is a `AuthFoo.from_dict` factory which generates it.

It is our responsibility as charm authors/maintainers to present a sane interface to charmed applications inside the boundaries of OF/charms. I would not want to force library users to go look at Grafana docs to determine what the arguments are supposed to be to construct a dict/yaml, or to use an abstracted `SomeConfigType` dataclass just to get default args/reasonable docstrings for an object which they will immediately pass.

Unless we get to a Spring-like level of abstraction (which I would hope we never do), I default to literacy and a sane user experience for library users over a cleaner experience for library maintainers/authors.

Even if this is pulled out into a "generalized" auth library used by more than one charm, I would probably still opt for something like:

```python
class GrafanaSomeAuthProvider(GrafanaAuthProvider): 
    def __init__(self, charm: CharmBase, args...):
        authConfig = AuthSomeMethod.from_dict(...)
        super(...)
```

We can (and should) "shield" library users from internals for default use cases, and simply make it easy to use.

```python
self.someauth = GrafanaAuthProvider(self, method=SomeOtherAuth(foo=bar, baz=quux), relation_name="some_non_default_relation", refresh_event=[self.on.pebble_ready, self.on.upgrade_charm])
```

Where Hawking joked/bet that the event horizon for singularities was because they were "obscene" and needed to be "decently covered", so too should the internals of making a charm "tick".

Make hard things easy, and make impossible things hard. That's the job for charming, and as elegant as the suggested constructor may be for us, as library authors and (relative) experts on how it operates, charming, and the Juju ecosystem, it's the sort of thing which an external contributor who wants to charm something and finds a library for auth may find as a very large hurdle without that familiarity.

lib/charms/grafana_auth/v0/grafana_auth.py
rbarry82: Realistically, the name of the class doesn't really matter, since the library consumer will only see the events from the library instance anyway.

However, I wonder if this should not be `AuthProxyProviderCharmEvents`, since other providers may not (and probably will not) need to sent exactly the same events.

rbarry82 (approved):

lib/charms/grafana_auth/v0/grafana_auth.py
saltiyazan: I don't think it is closely related to `AuthProxy`, it's more about the provider getting to know how to reach `Grafana`.
I see that some providers might not need the url, but those provider would not need to observe the event `urls_available`.
But actually `GitHub Auth` might need the url(s) too, as in described [Configure GitHub OAuth application](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github/#configure-github-oauth-application)
Furthermore, the [spec](https://docs.google.com/document/d/1OK8Ee4NtPbapMX2EKMLT_5bUnSWF0NPBYFQ5m9WEEKQ/edit#heading=h.ijcen96iw27) of the interface itself was that one side would provide the authentication configuration while the other side would provide the urls to access grafana.

lib/charms/grafana_auth/v0/grafana_auth.py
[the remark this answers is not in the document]
saltiyazan: I removed at the end.
The input should be validated by the lib :)
The jsonschema adds an extra check for the developer working on the lib and helps in having a more readable format instead of python variables. But the requirer or the provider shouldn't be able to break it for now, as it can only pass specific parameters.

rbarry82:
@saltiyazan Finally merged 👍

saltiyazan:
> @saltiyazan Finally merged 👍

Great news, thank you very much :)

PR #118, opened 2022-09-05
change  unit name

beliaev-maksim:
would it make sense to change a unitname according to the warning per https://github.com/canonical/operator/pull/822 ?

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -310,7 +310,7 @@ class TestCharmReplication(unittest.TestCase):
         self.harness.set_leader(True)
         self.harness.charm.on.config_changed.emit()
         rel = self.harness.model.get_relation("grafana")
-        self.harness.add_relation_unit(rel.id, "grafana/1")
+        self.harness.add_relation_unit(rel.id, "grafana-k8s/1")
 
         unit_ip = str(self.harness.charm.model.get_binding("grafana").network.bind_address)
         replica_address = self.harness.charm.get_peer_data("replica_primary")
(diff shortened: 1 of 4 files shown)


sed-i (commented):
Hi @beliaev-maksim, thanks for the PR!
I still see other additional tests fail on my end.
Could you please (temporarily) add this change to reqs txt so we can have CI reflect it?
```diff
-ops
+git+https://github.com/canonical/operator#egg=ops
```

beliaev-maksim:
you mean integrations that are not part of tox?
image


I ran tests on my fork, looks OK except libs check, but I assume because it is a fork
https://github.com/beliaev-maksim/grafana-k8s-operator/pull/1

sed-i:
I mean unit tests.
Can you please change `requirements.txt` so that `ops` is taken from gh instead of from pypi?
```diff
-ops
+git+https://github.com/canonical/operator#egg=ops
```

beliaev-maksim:
@sed-i 
it is in there:
image

or do you mean something else?

sed-i:
I mean unit tests.
Can you please change `requirements.txt` so that `ops` is taken from gh instead of from pypi?
```diff
-ops
+git+https://github.com/canonical/operator#egg=ops
```

sed-i:
Ok great, missed it.

beliaev-maksim:
@sed-i 
about "check libs". It looks like some issue in pipeline action itself.
even bumping it to stable via #119 get it failed.
credentials/secrets ?

tests/unit/test_charm.py
```diff
@@ -310,7 +310,7 @@ def test_primary_sets_correct_peer_data(self, mock_unit_ip):
         self.harness.set_leader(True)
         self.harness.charm.on.config_changed.emit()
         rel = self.harness.model.get_relation("grafana")
-        self.harness.add_relation_unit(rel.id, "grafana/1")
+        self.harness.add_relation_unit(rel.id, "grafana-k8s/1")
```
rbarry82, line 313: This will cause inconsistencies.

rbarry82 (changes requested):

simskij:
I'm happy to merge with check libs failing, if everything else is sorted out. I do, however, think Ryan is right and that we need to sort out [this](https://github.com/canonical/grafana-k8s-operator/pull/118#issuecomment-1238085615) first

rbarry82 (approved):

rbarry82:
Raised on MM for ops to look at, and the follow-up which rationalizes things LGTM.

simskij:
Thanks again @beliaev-maksim! 🎉

PR #119, opened 2022-09-06
bump all CI actions

beliaev-maksim:

PR #120, opened 2022-09-06
add a check for data sources

beliaev-maksim:
based on the CI fail:
https://github.com/canonical/grafana-k8s-operator/runs/8205066126?check_suite_focus=true

raise more explicit exception

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -127,7 +127,13 @@ def get_datasource_for(suffix: str, datasources: list) -> dict:
     Returns:
         a datasource config dict
     """
-    return [d for d in datasources if d["name"].endswith(suffix)].pop()
+    assert datasources, "'datasources' argument cannot be empty"
+
+    datasource_filtered = [d for d in datasources if d["name"].endswith(suffix)]
+    if not datasource_filtered:
+        raise ValueError("No data source was found for suffix {}".format(suffix))
+
+    return datasource_filtered.pop()
 
 
 async def get_grafana_dashboards(ops_test: OpsTest, app_name: str, unit_num: int) -> list:


rbarry82:
Let's put this into the PR in the comment. Spreading it across and linking to CI runs from other PRs hard to track back and forth

beliaev-maksim:
hmm, PRs are completely unrelated, except for the error message

and this PR will not fix a pipeline issue. It will be just a more explicit error, while the real root cause why datasource is empty is unclear

rbarry82:
> hmm, PRs are completely unrelated, except for the error message
> 
> and this PR will not fix a pipeline issue. It will be just a more explicit error, while the real root cause why datasource is empty is unclear

I can dig into the datasource issue, which may somehow be timing related with sqlite replication.

I meant that, in general, it's cleaner/easier for reviews and general process to have them all in one PR. In particular:

> based on the CI fail:
> https://github.com/canonical/grafana-k8s-operator/runs/8205066126?check_suite_focus=true

Is hard for reviewers. It's not from this PR, the run may be cleaned up at some point and can no longer be referenced, fixing it in a different PR wouldn't actually turn CI in the other green unless it was rebased after a merge, etc.

Can we merge these together?

beliaev-maksim:
@rbarry82 
my motivation was that each PR should be as small as possible and at the same time self consistent and represent a complete feature/bug fix.
logically, these two PRs are not related. One is about version bumping and another is about error msges

but I can glue them together

beliaev-maksim:
@rbarry82 
#119 was closed due to CI consolidation. That is actually an example why I split features in different PRs.

then I keep current PR opened

rbarry82:
> @rbarry82 #119 was closed due to CI consolidation. That is actually an example why I split features in different PRs.
> 
> then I keep current PR opened

Kind of the point is that the PRs _are_ logically related and are not self-consistent if one of them cannot pass CI without a change in a different PR.

beliaev-maksim:
@rbarry82 
I see what you mean

but unfortunately none of proposed by me PRs does not fix the pipeline, deeper investigation is required since now CI is failing even on `main`

current PR (#120) addresses just messaging for the future but does not fix the root cause

simskij (approved):

PR #121, opened 2022-09-07
Update unit name in tests

swalladge:
Name validation was recently added to add_relation_unit ([here]),
so we must update the remote unit name used here.
Besides that, it's probably best to be consistent with naming here - I guess there was a recent change from grafana to grafana-k8s, so this would have been easily missed.

Note that unittests were passing here,
but would otherwise fail once `ops` released a new update.

[here]: https://github.com/canonical/operator/pull/795/files#diff-37e8d80748bba21bb08c982f17b7879059bacc8d413b59da998a896fca18be33R712-R717

## Issue

Resolving this test failure on operator: https://github.com/canonical/operator/runs/8109846894  (`operator` has tests that run unit tests on selected charms, with the `ops` dependency installed at the latest revision) 

## Solution

see description above

## Testing Instructions

- install ops latest master from git: `git+https://github.com/canonical/operator/#egg=ops`
- run `tox -e unit`

This should fail on grafana-k8s-operator master, but pass on this branch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ jinja2 < 3
 kubernetes
 lightkube >= 0.11
 markupsafe == 2.0.1
-ops
+git+https://github.com/canonical/operator/#egg=ops
 pyyaml
 urllib3
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -310,7 +310,7 @@ class TestCharmReplication(unittest.TestCase):
         self.harness.set_leader(True)
         self.harness.charm.on.config_changed.emit()
         rel = self.harness.model.get_relation("grafana")
-        self.harness.add_relation_unit(rel.id, "grafana/1")
+        self.harness.add_relation_unit(rel.id, "grafana-k8s/1")
 
         unit_ip = str(self.harness.charm.model.get_binding("grafana").network.bind_address)
         replica_address = self.harness.charm.get_peer_data("replica_primary")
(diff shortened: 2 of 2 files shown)


dstathis (approved):

sed-i:
Hi @swalladge 
operator/795 is [not included](https://github.com/canonical/operator/compare/1.5.2...main) in ops 1.5.2.
Can you please (temporarily) update requirements.txt so ops is taken from github, to make sure CI passes?
```diff
-ops
+git+https://github.com/canonical/operator#egg=ops
```

Also, have you seen #118? Seems to overlap a bit.

beliaev-maksim:
@swalladge 
how does current PR differ from #118  ?

see comments in #118

rbarry82 (changes requested):
How does this differ from #118? It appears to be exactly the same, which means this comment is also applicable here. There also was no change from grafana to grafana-k8s.

This change is incomplete. 

In particular, the only place where the change from `ops` applies is in the naming of units in peer relations, which is present at least [here](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/unit/test_charm.py#L121), [here](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/unit/test_charm.py#L290), [here](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/unit/test_source_consumer.py#L80), and [here](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/unit/test_dashboard_consumer.py#L272). 

The current changes will actually make it more problematic, since the changed unit name (`grafana` -> `grafana-k8s`) will not actually match the name added to the relation in `setUp`, and we'll end up with a relation like:
```
{
  "grafana": [Unit<"grafana/0", ...>, Unit<"grafana-k8s/1", ...>]
}
```

We also intentionally avoid unreleased/from git dependencies where possible. Why re-add it?

simskij:
Thank you for your contributions, @swalladge! 🎉 I'll close this PR in favor of #118.

As a suggestion for next time, consider opening the PR against the branch of the pre-existing PR to have your changes merged there rather than to main. Just to ensure we don't end up with multiple PRs addressing the same issue. 👍🏼

PR #123, opened 2022-09-08
use binary to get workload version

dstathis:
## Issue
setting workload version could fail if Grafana is not ready yet.

## Solution
Run the binary to get the version.

## Testing Instructions
Check workload version is set.

## Release Notes
grafana-k8s sets workload version more reliably

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -22,6 +22,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import secrets
 import socket
 import string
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -746,11 +752,21 @@ def _build_replication(self, primary: bool) -> Layer:
...
+        container = self.containers["workload"]
+        if not container.can_connect():
+            return None
+        version_output, _ = container.exec(["grafana-server", "-v"]).wait_output()
+        # Output looks like this:
+        # Version 8.2.6 (commit: d2cccfe, branch: HEAD)
+        result = re.search(r"Version (\d*\.\d*\.\d*)", version_output)
+        if result is None:
+            return result
+        return result.group(1)
```
sed-i, line 769: This is typically written with a walrus:
```suggestion
        if match := re.search(r"Version (\d*\.\d*\.\d*)", version_output):
            return match.group(1)
        return None
```

sed-i (approved):

src/charm.py
rbarry82: As said many times before, we should avoid this if possible. That would be literally the only occurrence of a walrus inside the Grafana codebase.

It costs us very little and potentially benefits quite a bit (similar to the request for a machine version of the `grafana-agent` charm, which we cannot guarantee will run on a recent version of Python) if a scenario comes up where the charm codebase also needs to run with older versions.

dstathis: Yeah I was avoiding based on @rbarry82's argument.

PR #124, opened 2022-09-13
Add self-monitoring dashboard

### LINKED ISSUE #97: Add alert rules, relations and dashboards for self-monitoring

rbarry82:
## Issue
Self-monitoring is already present, but add a self-monitoring dashboard. Partially closes #97 

## Solution
Grafana 9 (and Grafana Cloud) have their own monitoring store, but it's not exposed in Grafana 8 yet.

Instead, once Grafana is connected to a Prometheus, a dashboard can be imported from the UI. This expects Grafana to have the `/metrics` endpoint enabled _and_ that that Prometheus is scraping it.

This is... messy from a charm perspective. So we need to check whether there is a `prometheus_scrape` _and_ a `grafana_source` to the same application. If that's true, then we can provision our own. But we cannot use the library without a major refactor which discards all of the relation validation or provides a back-channel, and the majority of what the library provides (dropdown injection, dynamic topology, validation/messaging) is not useful anyway.

So push it into the container directly if the conditions are met.

## Release Notes
Add self-monitoring dashboard

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -220,7 +220,46 @@ class GrafanaCharm(CharmBase):
         Args:
             event: a :class:`GrafanaSourceEvents` instance sent from the provider
         """
+        self._maybe_provision_own_dashboard()
         self._configure()
+
+    def _maybe_provision_own_dashboard(self) -> None:
+        """If all the prerequisites are enabled, provision a self-monitoring dashboard.
+
+        Requires:
+            A Prometheus relation on self.prometheus_scrape
+            The SAME Prometheus relation again on grafana_source
+
+        If those are true, we have a Prometheus scraping this Grafana, and we should
+        provision our dashboard.
+        """
+        container = self.unit.get_container(self.name)
+        if not container.can_connect():
+            logger.debug("Cannot connect to Pebble yet, not provisioning own dashboard")
+            return
+
+        source_related_apps = self.model.relations["grafana-source"]
+        scrape_related_apps = self.model.relations["metrics-endpoint"]
+
+        has_relation = any(
+            [source for source in source_related_apps if source in scrape_related_apps]
+        )
+
+        dashboards_dir_path = os.path.join(PROVISIONING_PATH, "dashboards")
+        self.init_dashboard_provisioning(dashboards_dir_path)
+
+        dashboard_path = os.path.join(dashboards_dir_path, "grafana_metrics.json")
+        if has_relation and self.unit.is_leader():
+            # This is not going through the library due to the massive refactor needed in order
+            # to squash all of the `validate_relation_direction` and structure around smashing
+            # the datastructures for a self-monitoring use case. This is built-in in Grafana 9,
+            # so it will FIXME be removed when we upgrade
+            container.push(dashboard_path, "src/self_dashboard.json", make_dirs=True)
+        elif not has_relation and "grafana_metrics.json" in container.list_files(
+            dashboards_dir_path, pattern="grafana_metrics.json"
+        ):
+            container.remove_path(dashboard_path)
+            logger.debug("Removed dashboard %s", dashboard_path)
 
     def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
         """Re-provision Grafana and its datasources on upgrade.
--- a/src/charm.py
+++ b/src/charm.py
@@ -50,6 +50,7 @@ from ops.charm import (
     ActionEvent,
     CharmBase,
     ConfigChangedEvent,
+    HookEvent,
     RelationBrokenEvent,
     RelationChangedEvent,
     UpgradeCharmEvent,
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -220,8 +220,47 @@ def _on_grafana_source_changed(self, _: GrafanaSourceEvents) -> None:
         Args:
             event: a :class:`GrafanaSourceEvents` instance sent from the provider
         """
+        self._maybe_provision_own_dashboard()
```
sed-i, line 223: - Should this also be called on metrics endpoint changed?
- Since this call is not inside _configure, do we need additional "recovery" mechanisms in case it didn't get provisioned the first time?

src/charm.py
```diff
@@ -220,8 +220,47 @@ def _on_grafana_source_changed(self, _: GrafanaSourceEvents) -> None:
...
+        self.init_dashboard_provisioning(dashboards_dir_path)
+
+        dashboard_path = os.path.join(dashboards_dir_path, "grafana_metrics.json")
+        if has_relation and self.unit.is_leader():
+            # This is not going through the library due to the massive refactor needed in order
+            # to squash all of the `validate_relation_direction` and structure around smashing
+            # the datastructures for a self-monitoring use case. This is built-in in Grafana 9,
+            # so it will FIXME be removed when we upgrade
+            container.push(dashboard_path, "src/self_dashboard.json", make_dirs=True)
+        elif not has_relation and "grafana_metrics.json" in container.list_files(
```
sed-i, line 258: The `list_files` check seems unnecessary given we have the `rm -rf` equivalent: `remove_path(..., recursive=True)`
https://github.com/canonical/operator/blob/5bc3acb40fc0db4887b2377fd0ed7a9591df78e9/ops/pebble.py#L2102

src/charm.py
```diff
@@ -220,8 +220,47 @@ def _on_grafana_source_changed(self, _: GrafanaSourceEvents) -> None:
...
+
+        source_related_apps = self.model.relations["grafana-source"]
+        scrape_related_apps = self.model.relations["metrics-endpoint"]
+
+        has_relation = any(
+            [source for source in source_related_apps if source in scrape_related_apps]
+        )
+
+        dashboards_dir_path = os.path.join(PROVISIONING_PATH, "dashboards")
+        self.init_dashboard_provisioning(dashboards_dir_path)
```
sed-i, line 249: Shouldn't we restart grafana only after the file is written/removed?

rbarry82: See [here](https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards).

> When Grafana starts, it will update/insert all dashboards available in the configured path. Then later on poll that path every updateIntervalSeconds and look for updated json files and update/insert those into the database.

We do not need to restart Grafana.

src/charm.py
rbarry82: I'd rather scalpel than hammer, but I'm not particular about it.

sed-i: Seems like the less potential pebble errors the better.
If this raises an exception on shutdown then the unit/app won't ever be removed, right?

dstathis: Wouldn't the remove_path raise an exception in that case anyway?

src/charm.py
```diff
@@ -220,8 +220,47 @@ def _on_grafana_source_changed(self, _: GrafanaSourceEvents) -> None:
...
+            A Prometheus relation on self.prometheus_scrape
+            The SAME Prometheus relation again on grafana_source
+
+        If those are true, we have a Prometheus scraping this Grafana, and we should
+        provision our dashboard.
+        """
+        container = self.unit.get_container(self.name)
+        if not container.can_connect():
+            logger.debug("Cannot connect to Pebble yet, not provisioning own dashboard")
+            return
```
dstathis, line 253: If this can fire before pebble ready, shouldn't we run it in pebble_ready as well to guarantee it gets run?

src/charm.py
rbarry82: Yes, it would raise an exception anyway, but there's a guard at the top of the method to ensure we can actually connect.

src/charm.py
rbarry82: It's more that it can fire on `RelationBroken` if a model is being torn down. It has a guaranteed run due to `grafana_source` handling `upgrade_keys` (to move from `StoredState` to peer databags) on upgrade, and the same observer sends it here.

tests/integration/test_selfmonitoring_dashboard.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
```
sed-i, line 4: Would you mind adding a module level docstring for overall description?

sed-i (approved):

Abuelodelanada (approved):

PR #129, opened 2022-09-21
update from landing-page-k8s to catalogue-k8s

simskij:
Update to reflect the renaming of landing-page-k8s to catalogue-k8s

### DIFF
--- a/lib/charms/catalogue_k8s/v0/catalogue.py
+++ b/lib/charms/catalogue_k8s/v0/catalogue.py
@@ -1,39 +1,44 @@
-"""Charm for providing landing pages to bundles."""
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Charm for providing services catalogues to bundles or sets of charms."""
 
 import ipaddress
-import socket
-from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents
-from ops.charm import CharmBase
-
 import logging
+import socket
+from typing import List, Optional, Union
 
-LIBID = "7e5cd3b1e1264c2689f09f772c9af026"
+from ops.charm import CharmBase
+from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents
+
+LIBID = "fa28b361293b46668bcd1f209ada6983"
 LIBAPI = 0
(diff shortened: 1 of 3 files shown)


lib/charms/catalogue_k8s/v0/catalogue.py
```diff
@@ -1,39 +1,44 @@
...
+    """`CatalogueConsumer` is used to send over a `CatalogueItem`."""
 
     def __init__(
-        self, 
+        self,
         charm,
-        relation_name: str = DEFAULT_RELATION_NAME, 
-        app: LandingPageApp = None
+        relation_name: str = DEFAULT_RELATION_NAME,
+        app: CatalogueItem = None,
```
sed-i, line 40: Seems like we should change `app` to `item`/`entry`/etc.

lib/charms/catalogue_k8s/v0/catalogue.py
```diff
@@ -1,39 +1,44 @@
...
 logger = logging.getLogger(__name__)
 
-class LandingPageApp:
-    name: str
-    url: str
-    icon: str
-    description: str
 
-    def __init__(self, name, url, icon, description = ""):
+class CatalogueItem:
```
sed-i, line 23: On one hand, this could be a [TypedDict](https://www.python.org/dev/peps/pep-0589/).
On the other hand, since `app=None` has a default value, this class could have a `.from_charm` utility.

WDYT?

sed-i (approved):

lib/charms/catalogue_k8s/v0/catalogue.py
simskij: Sure, let's do that before merging this.

Abuelodelanada (approved):

dstathis (approved):

metadata.yaml
```diff
@@ -41,8 +41,8 @@ requires:
   database:
     interface: db
     limit: 1
-  landing-page:
-    interface: catalog
+  catalogue:
+    interface: catalogue
```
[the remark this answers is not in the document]
simskij: I looked it up, and apparently catalogue is the most commonly used spelling outside of the US. 😅

PR #135, opened 2022-10-03
Move the ingress relation under requires

sed-i:
## Issue
Cannot relate grafana to traefik.

## Solution
- Move the ingress relation under requires
- Add itest

## Context
```
$ juju relate trfk graf
ERROR no relations found
```

## Testing Instructions
Relate grafana to traefik.

## Release Notes
Move the ingress relation under requires


simskij (approved):

src/charm.py
```diff
@@ -1069,11 +1069,11 @@ def external_url(self) -> str:
...
 
     @property
-    def _ingress_config(self) -> str:
+    def _ingress_config(self) -> dict:
         """Build a raw ingress configuration for Traefik."""
         fqdn = urlparse(socket.getfqdn()).path
-        domain = fqdn.split(
-            "{}.{}.".format(self.model.unit.name.replace("/", "-"), self.model.name)
+        domain = re.split(
+            r"^{}\..*?{}\.".format(self.model.unit.name.replace("/", "-"), self.model.name), fqdn
```
Abuelodelanada, line 1076: Reading regular expressions is not always easy.... 

What about this:

```python
pattern = r"^{}\..*?{}\.".format(self.model.unit.name.replace("/", "-"), self.model.name)
domain = re.split(pattern, fqdn)
```

Abuelodelanada (approved):

src/charm.py
sed-i: Fixed in next commit.

rbarry82: Either works for me, though the difference here mostly seems to be in assigning it as a separate variable. If we're going to do that, we may as well `re.compile()` while we're at it

rbarry82:
Let's merge as soon as CI passes

rbarry82 (approved):

sed-i:
A seemingly unrelated itest repeatedly fails now: #137

PR #146, opened 2022-10-18
Use cos-tool to template with Grafana variables instead of 'raw' topology

rbarry82:
## Issue
Currently, `cos-tool` via `grafana-dashboard` is injecting "actual"/"realized" topology into dashboard panels rather than template variables, which is undesirable.

## Solution
Remap the keys into variables before injection

## Release Notes
Use cos-tool to template with Grafana variables instead of 'raw' topology


simskij (approved):
Happy for this to be merged as soon as CI passes!

PR #152, opened 2022-10-26
Add more guards around relation properties

rbarry82:
## Issue
On `relation-broken` for the ingress relation, it seems to be possible to receive a valid `Relation` which has no `app` property set.

## Solution
Guard against non-truthy `relation.app`. In case this occurs, explicitly clear the stored `external_url` on `relation-broken`.

## Testing Instructions
Relate to grafana. Break the relation.

## Release Notes
Add more guards around relation properties in `TraefikRouteRequirer`

### DIFF
--- a/lib/charms/traefik_route_k8s/v0/traefik_route.py
+++ b/lib/charms/traefik_route_k8s/v0/traefik_route.py
@@ -213,6 +213,9 @@ class TraefikRouteRequirer(Object):
         self.framework.observe(
             self._charm.on[relation_name].relation_changed, self._on_relation_changed
         )
+        self.framework.observe(
+            self._charm.on[relation_name].relation_broken, self._on_relation_broken
+        )
 
     @property
     def external_host(self) -> str:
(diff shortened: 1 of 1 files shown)


sed-i (commented):
Could you please elaborate why `_stored` is central to this fix?

rbarry82:
`_stored` is not central to this fix, except in that `TraefikRoute` uses `_stored`, so the changes to it followed that.

Here, it is being explicitly cleared on `RelationBroken` or if `app` is falsy so we do not leave an "old" `external_host` which is not valid anymore (either because it's `RelationBroken` or because there is no `relation.app` during `relation-changed` during a break), and we reflect back to the charm that "there is no meaningful relation or the relation is breaking, so when `ingress.ready` is fired and `ingress.external_host` is checked, it returns a falsy value and the logic in the charm continues"

sed-i (approved):
Just realized that this is a fetch-lib PR!

PR #154, opened 2022-10-27
Make sure racy external_host returns from Traefik still set SERVE_FROM_SUB_PATH

rbarry82:
## Issue
On CPU-constrained systems, reconfiguration of Grafana may happen before `external_host` resolves the route from Traefik, leaving to a layer config which does not include `SERVE_FROM_SUBPATH`, and requests redirect to `http://uri/login` (for example)

## Solution
Catch `ingress.on.ready` and ensure the Grafana layer has the correct configuration

## Release Notes
Make sure racy external_host returns from Traefik still set SERVE_FROM_SUB_PATH

### DIFF
--- a/tests/unit/test_external_url.py
+++ b/tests/unit/test_external_url.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+import unittest
+from typing import Dict
+from unittest.mock import MagicMock, patch
+
+import ops
+from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
+from ops.model import ActiveStatus
+from ops.testing import Harness
+
+from charm import PORT, GrafanaCharm
+
+logger = logging.getLogger(__name__)
+
+ops.testing.SIMULATE_CAN_CONNECT = True
+CONTAINER_NAME = "grafana"
+SERVICE_NAME = "grafana"
+
+k8s_resource_multipatch = patch.multiple(
+    "charm.KubernetesComputeResourcesPatch",
+    _namespace="test-namespace",
+    _patch=lambda *a, **kw: True,
+    is_ready=lambda *a, **kw: True,
+)
+
+
+class TestExternalUrl(unittest.TestCase):
+    """External url sources have a precedence, and must be correctly propagated everywhere."""
+
+    def setUp(self, *unused):
+        self.harness = Harness(GrafanaCharm)
+        self.addCleanup(self.harness.cleanup)
+
+        model_name = "testmodel"
+        self.harness.set_model_name(model_name)
+        for p in [
+            patch("charm.K8sServicePatch"),
+            patch("lightkube.core.client.GenericSyncClient"),
+            patch(
+                "socket.getfqdn", new=lambda *args: f"grafana-k8s-0.{model_name}.svc.cluster.local"
+            ),
+            patch("socket.gethostbyname", new=lambda *args: "1.2.3.4"),
+            k8s_resource_multipatch,
+            patch.object(GrafanaCharm, "grafana_version", "0.0.0"),
+            patch("ops.testing._TestingModelBackend.network_get"),
+            patch("ops.testing._TestingPebbleClient.exec", MagicMock()),
+        ]:
+            p.start()
+            self.addCleanup(p.stop)
+
+        self.harness.set_leader(True)
+
+        # Peer relation
+        self.app_name = "grafana-k8s"
+        self.peer_rel_id = self.harness.add_relation("grafana", self.app_name)
+
+        # Auth relation
+        # FIXME why tests fail when this relation is not created?
+        self.grafana_auth_rel_id = self.harness.add_relation("grafana-auth", "auth_provider")
+
+        self.harness.begin_with_initial_hooks()
+        self.harness.container_pebble_ready(CONTAINER_NAME)
+        self.harness.container_pebble_ready("litestream")
+        self.fqdn_url = f"http://fqdn:{PORT}"
+
+    def get_pebble_env(self) -> Dict[str, str]:
+        service = (
+            self.harness.get_container_pebble_plan(CONTAINER_NAME).services["grafana"].to_dict()
+        )
+        return service["environment"]
+
+    def is_service_running(self) -> bool:
+        service = self.harness.model.unit.get_container(CONTAINER_NAME).get_service(SERVICE_NAME)
+        return service.is_running()
+
+    @patch.object(TraefikRouteRequirer, "external_host", new="1.2.3.4")
+    def test_url_without_path(self):
+        """The root url and subpath env vars should not be set when no subpath is present."""
+        # GIVEN a charm with an fqdn as its external URL
+        # (this is set by a mock decorator)
+
+        # THEN root url and subpath envs are not defined (redirect to login is fine with a bare hostname)
+        self.assertNotIn("GF_SERVER_SERVE_FROM_SUB_PATH", self.get_pebble_env())
+        self.assertNotIn("GF_SERVER_ROOT_URL", self.get_pebble_env())
+        self.assertTrue(self.is_service_running())
+
+    def test_external_url_precedence(self):
+        """Precedence is [config option] > [ingress] > [fqdn]."""
+        # GIVEN a charm with the fqdn as its external URL
+        # (this is set by a mock decorator)
+
+        # WHEN a relation with traefik is formed
+        with patch.object(TraefikRouteRequirer, "external_host", new="1.2.3.4"):
+            rel_id = self.harness.add_relation("ingress", "traefik-app")
+            self.harness.add_relation_unit(rel_id, "traefik-app/0")
+
+            # AND ingress becomes ready
+            self.harness.charm.ingress.on.ready.emit(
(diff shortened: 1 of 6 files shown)


tests/unit/test_external_url.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Peer relation
+        self.app_name = "grafana-k8s"
+        self.peer_rel_id = self.harness.add_relation("grafana", self.app_name)
+
+        # Auth relation
+        # FIXME why tests fail when this relation is not created?
+        self.grafana_auth_rel_id = self.harness.add_relation(
+            "grafana-auth", "auth_provider"
+        )
```
sed-i, line 57: @rbarry82 could you please take a look at this FIXME?

tests/unit/test_external_url.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @patch.object(TraefikRouteRequirer, "external_host", new="1.2.3.4")
+    def test_external_url_precedence(self):
+        """Precedence is [config option] > [ingress] > [fqdn]."""
+        # GIVEN a charm with the fqdn as its external URL
+        # (this is set by a mock decorator)
+
+        # FIXME why the test fails without re-emitting leader elected here?
+        self.harness.set_leader(False)
+        self.harness.set_leader(True)
```
sed-i, line 99: @rbarry82 could you please take a look at this FIXME?

sed-i (commented):
I added a few utests and have some failures at the moment.
I am looking into this now but perhaps you'd find the root cause sooner.

sed-i:
@rbarry82 the remaining utest failure is for invalid url scenarios.
- When we set `htp:/foo.bar`, then `GF_SERVER_SERVE_FROM_SUB_PATH` is define
- When we set `htp://foo.bar`, then `GF_SERVER_SERVE_FROM_SUB_PATH` is NOT define

What's the intended behavior for invalid urls?

rbarry82:
> @rbarry82 the remaining utest failure is for invalid url scenarios.
> 
> * When we set `htp:/foo.bar`, then `GF_SERVER_SERVE_FROM_SUB_PATH` is defined
> * When we set `htp://foo.bar`, then `GF_SERVER_SERVE_FROM_SUB_PATH` is NOT defined
> 
> What's the intended behavior for invalid urls?

The admin intentionally set it. Leave it.

sed-i:
> The admin intentionally set it. Leave it.

Marked test with `skip`.

Remaining failure: when ingress relation is removed, envvars are not updated.

rbarry82:
> > The admin intentionally set it. Leave it.
> 
> Marked test with `skip`.
We should just delete dead code instead of committing it.

> 
> Remaining failure: when ingress relation is removed, envvars are not updated.

This is wrong. Breaking the relation [clears the stored external host and emits a ready event](https://github.com/canonical/grafana-k8s-operator/blob/main/lib/charms/traefik_route_k8s/v0/traefik_route.py#L249). The ready event is caught, `_configure` is invoked, which will [check `_build_layer`](https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L399), and `extra_info` [will never be updated](https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L815). 

This is likely a problem with the unit test, but as said earlier, since the essential parts of the unit tests here are already covered by two existing unit tests and tracing the code clearly demonstrates what the actual behavior is, I'll defer looking at the unit test until I make the integration test stable.

sed-i:
> We should just delete dead code instead of committing it.

We could, but invalid url seems like a solid use case. Letting the admin have it is fine, but it seems that charm code produces different results for different kinds of invalid output.

> This is likely a problem with the unit test

Indeed. Fixed.

rbarry82:
> We could, but invalid url seems like a solid use case. Letting the admin have it is fine, but it seems that charm code produces different results for different kinds of invalid intput.

Not per-se. `htp:/blah/blah` is clearly invalid, _but_ `urlparse` assumes that anything which does not start with an explicit scheme is just a path, and at best, it tries to get query params out from there. All of our charms which use `urlparse` (which is basically all of them) would do the same thing here. For example:

```python
>>> from urllib.parse import urlparse
>>> urlparse('htp:/foo/bar')
ParseResult(scheme='', netloc='', path='htp:/foo/bar', params='', query='', fragment='')
>>> urlparse('1.2.3.4/.foo/bar')
ParseResult(scheme='', netloc='', path='1.2.3.4/.foo/bar', params='', query='', fragment='')
>>> urllib.parse.urlparse('htp://foo/bar')
ParseResult(scheme='htp', netloc='foo', path='/bar', params='', query='', fragment='')
```

When there is `<scheme>://<address>/<path>?<query>=<params>`, `urlparse` doesn't try to be smart. It follows that anything before `://` is a scheme instead of matching it against a list, that any `<address>` inbetween `://` and the next `/` is a location, even if it's something like `1.2.3.4.5.6.7` (which is potentially a valid DNS record, sadly, but kind of to the point that it doesn't even bother checking whether it could be an IP or trying to resolve it), and so on.

It's relatively naive, and the fact that `urlparse` thinks that `🤷‍♂️://0xDEADBEEF/™` has appropriate scheme/location/path is one thing. The charm is somewhat relying (like all our charms) on the fact that, if `urlparse().path` is truthy, we should trust it. If we're going to try to validate/sanitize input, we should do it in all of our charms, and it should be caught with tests/code _way before_ we get to "is the environment variable set", in something more like "the charm is blocked because the `config` value you set doesn't make sense"

sed-i:
> we should do it in all of our charms

Agreed.
We do have some attempts in some places but we should certainly be more consistent about it.

- https://github.com/canonical/alertmanager-k8s-operator/blob/e2ef70b353fc48bc55405cdf7bff25d670b24365/src/charm.py#L557
- https://github.com/canonical/prometheus-k8s-operator/blob/4e15dc298948e20877c715ea47715b750198e1ee/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L577

After all, blocking on key input being invalid is probably expected.

sed-i:
Utests pass now in CI but for some reason `test_dashboard_transform.py::TestDashboardLabelInjector` fail on mac:

```
FAILED tests/unit/test_dashboard_transform.py::TestDashboardLabelInjector::test_consumer_adds_labels - AssertionError: Lists differ: [{'id': 'file:tester', 'relation_id': '2', 'charm': 'grafana-k8s', 'content': '{"panels": [{"data": "label_values(up, juju_unit)", "datasourc...
FAILED tests/unit/test_dashboard_transform.py::TestDashboardLabelInjector::test_consumer_adds_labels_for_loki - AssertionError: Lists differ: [{'id': 'file:tester', 'relation_id': '2', 'charm': 'grafana-k8s', 'content': '{"panels": [{"data": "label_values(up, juju_unit)", "datasourc...
FAILED tests/unit/test_dashboard_transform.py::TestDashboardLabelInjector::test_consumer_handles_expressions_with_offsets - AssertionError: Lists differ: [{'id': 'file:tester', 'relation_id': '2', 'charm': 'grafana-k8s', 'content': '{"panels": [{"data": "label_values(up, juju_unit)", "targets":...
FAILED tests/unit/test_dashboard_transform.py::TestDashboardLabelInjector::test_consumer_handles_expressions_with_ranges - AssertionError: Lists differ: [{'id': 'file:tester', 'relation_id': '2', 'charm': 'grafana-k8s', 'content': '{"panels": [{"data": "label_values(up, juju_unit)", "targets":...
FAILED tests/unit/test_dashboard_transform.py::TestDashboardLabelInjector::test_consumer_handles_negatives_and_multiple_targets - AssertionError: Lists differ: [{'id': 'file:tester', 'relation_id': '2', 'charm': 'grafana-k8s', 'content': '{"panels": [{"data": "label_values(up, juju_unit)", "targets":...
```

rbarry82:
I'm ok with them failing on Macs. `DeepDiff` would probably resolve it, but guessing at the relative in-memory structure of dicts on a different and unsupported operating system now that dicts are ordered may not be worth it.

Similarly, considering that `web_external_url` is primarily used for administrators to specify an oauth/oidc callback address, and is only in Grafana to make load testing easier before ingress was added (as we have no IAM story at the moment), I'd just as soon remove the option entirely if possible until we do than try to guess at what is and is not valid, since the only side effect is _if_ an "invalid" `external_url` is intentionally specified, than an environment variable may or may not be set in unit tests.

None of the other places have the same usage as Grafana does, as they do not have an equivalent to `SERVE_FROM_SUB_PATH`. Alertmanager is always `http[s]`, and the HTTP handler is rooted there (same for Prometheus). These aren't the same. Really. Neither of those (and none of our other charms) have the potential for oauth2 callbacks as part of the application. They all need an external proxy wired in if we implement it.

It is, functionally, equivalent to stabbing ourselves in the eye and complaining that it hurts. Attempts to "validate" the correctness of it without any spec (since there's no IAM story) potentially breaks any hypothetical users who are relying strange callback addresses and serves no benefit other than a unit test which is written with assumptions that are not in any spec anywhere. 

Until such time as there is a spec, what's currently in [`config.yaml`](https://github.com/canonical/grafana-k8s-operator/blob/main/config.yaml#L25) is more or less guidance-only, and "should be..." isn't something we need to write set-in-stone unit tests about.

sed-i:
> fail on mac

Apparently I was hallucinating - they do not pass CI. DeepDiff doesn't help.

Abuelodelanada (approved):

Issue #137, opened 2022-10-04
Flaky integration tests


sed-i:
### Bug Description
`test_selfmonitoring_dashboard::test_remove` occasionally fails.

After removing prom,
```python
await ops_test.model.applications[prometheus_app_name].remove()
```

The test makes an assertion
```python
relation_removed_dashboards = await get_grafana_dashboards(ops_test, grafana_app_name, 0)
assert relation_removed_dashboards == []
```

That fails. Occasionally.
https://github.com/canonical/grafana-k8s-operator/actions/runs/3177380221/jobs/5177760631#step:5:801

### To Reproduce

Run itests.

### Environment

GitHub runner

### Relevant log output

```shell
Traceback (most recent call last):
  File "/home/runner/work/grafana-k8s-operator/grafana-k8s-operator/tests/integration/test_selfmonitoring_dashboard.py", line 88, in test_remove
    assert relation_removed_dashboards == []
AssertionError: assert [{'id': 1, 'i...eta': 0, ...}] == []
  Left contains one more item: {'id': 1, 'isStarred': False, 'slug': '', 'sortMeta': 0, ...}
  Full diff:
    [
  -  ,
  +  {'id': 1,
  +   'isStarred': False,
  +   'slug': '',...
```

sed-i:
Seems like it fails more frequently now
https://github.com/canonical/grafana-k8s-operator/actions/runs/3184129557/jobs/5199389512#step:5:809

rbarry82:
This was actually flaky before, and to make it less flaky, removing the "self-monitoring" relation actually [completely restarts Grafana after deleting the provisioning file](https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L342). I'll try with microk8s 1.25 locally to see if I can make this happen more often locally.

It's completely crazy that the `ModelConfigChange` workaround [here](https://github.com/canonical/grafana-k8s-operator/pull/135/commits/b226268712ac37a24d271e71b1c93e65491b6031) increased reliability, because Grafana doesn't use update-status... at all. So that change is affecting Juju behavior only, yet increases reliability of tests in a charm which literally do not respond to that event.

sed-i:
Still flaky:

- `test_selfmonitoring_dashboard`: [still fails](https://github.com/canonical/grafana-k8s-operator/actions/runs/3492470981/jobs/5854092283).
[...]

PR #156, opened 2022-11-17
fetch-lib

sed-i:
## Issue
Exception on ingress removal.

## Solution
Fetch lib.


lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, *args):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 2
```
Abuelodelanada, line 91: Are we downgrading the lib?

sed-i:
charmcarft says:
```
Library charms.grafana_auth.v0.grafana_auth has local changes, cannot be updated.
Library charms.grafana_k8s.v0.grafana_auth was already up to date in version 0.1.
```
charm code only uses
```
from charms.grafana_auth.v0.grafana_auth import AuthRequirer, AuthRequirerCharmEvents
```

Both files have the same LIBID.

Any cleanup to do here? @rbarry82 @saltiyazan

lib/charms/traefik_route_k8s/v0/traefik_route.py
sed-i: I think it was incremented without publishing.

saltiyazan (approved):

sed-i:
CI fails for reasons documented in #137. Merging.

PR #158, opened 2022-11-22
More refresh events for catalogue

sed-i:
## Issue
In a load test deployment, the link was still wrong.

## Solution
Add more refresh events.

## Context
Addresses https://github.com/canonical/catalogue-k8s-operator/issues/3.

## Testing Instructions
1. Deploy the cos-lite bundle.
2. Make sure grafana has the correct url.

## Release Notes
Add more refresh events for catalogue.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -203,7 +203,10 @@ class GrafanaCharm(CharmBase):
             refresh_event=[
                 self.on.grafana_pebble_ready,
                 self.ingress.on.ready,
-                self.ingress.on.revoked,
+                # TODO use revoked instead of relation events when it becomes available
+                # https://github.com/canonical/traefik-route-k8s-operator/issues/21
+                # self.ingress.on.revoked,
+                self.on["ingress"].relation_broken,
                 self.on.config_changed,  # web_external_url; also covers upgrade-charm
                 # TODO remove the following after the traefik issue is fixed
                 #  https://github.com/canonical/traefik-k8s-operator/issues/78
--- a/src/charm.py
+++ b/src/charm.py
@@ -208,11 +208,6 @@ class GrafanaCharm(CharmBase):
                 # self.ingress.on.revoked,
                 self.on["ingress"].relation_broken,
                 self.on.config_changed,  # web_external_url; also covers upgrade-charm
-                # TODO remove the following after the traefik issue is fixed
-                #  https://github.com/canonical/traefik-k8s-operator/issues/78
-                self.on["ingress"].relation_joined,
-                self.on["ingress"].relation_changed,
-                self.on.update_status,
             ],
             item=CatalogueItem(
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -202,8 +202,17 @@ def __init__(self, *args):
             charm=self,
             refresh_event=[
                 self.on.grafana_pebble_ready,
+                self.ingress.on.ready,
+                # TODO use revoked instead of relation events when it becomes available
+                # https://github.com/canonical/traefik-route-k8s-operator/issues/21
```
rbarry82, line 207: `traefik-route` should never handle this. Grafana is a unique use case which is somewhat abusing the interface/library, and the traefik route charm itself handles IPU and IPA when using the lib. We don't need this TODO. 

In addition, `["ingress"].changed` more or less covers `ingress.on.ready`

sed-i: > traefik-route should never handle this.

I don't follow. The charm is using `self.ingress = TraefikRouteRequirer(...`. Shouldn't `TraefikRouteRequirer` observe relation-departed and emit a `revoked`?

> `["ingress"].changed` more or less covers `ingress.on.ready`

Observing the `changed` ourselves may create a race where a hypothetical e.g. `ingress.url` is not yet set, because the lib processes the `changed` later. But we have an established pattern where libs hide relation events as much as possible.

rbarry82: `TraefikRouteRequirer` was pulled out of `traefik-route-k8s`, which is structued more like `prometheus-scrape-config`. `TraefikRoute` is intended to _modify_ IPU/IPA relations, which `traefik-route` uses. It was just a convenient mechanism for Grafana (which has a fairly unique "IPU, but only for the leader, and the URL shouldn't change when the leader does" use case, which we could use the lib for to send a "raw" traefik config). It isn't intended to be a general-consumption library which would follow those patterns.

Three's no possible race in observing the change. See [here](https://github.com/canonical/grafana-k8s-operator/blob/main/lib/charms/traefik_route_k8s/v0/traefik_route.py#L197). The only time `TraefikRouteProvider` "changes" the relation at all is in response to `TraefikRouteRequirer` (Grafana) sending config data, or some other event on the Traefik side which may change _its_ external host. 

In turn, either `relation-changed` or `relation-broken` will emit a `ready` event (so we didn't change the API of the library too much just for the Grafana use case). The libraries aren't hiding relation _events_ as much as possible so much as hiding interacting with relation _data_ as much as possible. For `catalogue-k8s` explicitly watching for events to refresh itself, there's not a meaningful difference to me. Either works, but we should stick to one of the other (`ingress.on.ready`, which covers both cases, or `["ingress"].changed && ["ingress"].broken`) rather than mixing metaphors.

rbarry82 (approved):

PR #159, opened 2022-11-29
Don't inject labels with cos-tool if dropdown injection is disabled

rbarry82:
## Issue
When injection with `cos-tool` was added, even if `inject_dropdowns` is false (for "static" dashboards via `cos-config` or other), the addition of topology in the provided relation data will trigger injection via cos-tool

## Solution
If `inject_dropdowns` is `True`, also avoid passing along toplogy.

## Release Notes
Don't inject labels with cos-tool if dropdown injection is disabled

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -218,7 +218,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sed-i (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1155,7 +1155,7 @@ def _content_to_dashboard_object(self, content: str, inject_dropdowns: bool = Tr
         return {
             "charm": self._charm.meta.name,
             "content": content,
-            "juju_topology": self._juju_topology,
+            "juju_topology": self._juju_topology if inject_dropdowns else {},
```
PietroPasotti, line 1158: To someone new reading this (like me), it's not obvious what's the reason for this conditional.
Juju topology is empty if inject_dropdowns is true?

It may be cleaner code if we found a different way to pass down `inject_dropdowns` as a value to whoever is the consumer of the return value of this 'dashboard object', so that they can do what's appropriate in a more semantically transparent way.

That, or add a comment explaining what's going on :)

PietroPasotti (approved):
LGTM, modulo the comment about the opaque code

PR #185, opened 2023-01-31
add All option and multi-select to dashboard variables

### LINKED ISSUE #145: Add the "All" option for Juju Unit in the Dashboard Filters
### Enhancement Proposal

Some dashboards would likely want to be able to use juju topology down to the application level, but omit the unit selector, for instance to create lists of units and their statuses. Currently, that option is missing.

lucabello:
## Issue
Closes #145.

## Solution
Set the flags to enable "All" and multi-select to True. It was requested to do this for all the `juju_*` variables and for the datasources.


rbarry82:
Other than the unit tests, LGTM. It's currently `allValue: None`, which I _think_ should interpolate to "ignore this variable if "All" is selected", but verification would be great.

lucabello:
@rbarry82 `allValue: None` means we are not specifying a "custom value" for the "All" option; it allows you to configure the "All" option to be more restrictive, but we don't need that :) 
In any case I did double check, and it's working fine! 

I'll fix the unit tests, as I forgot to update them to follow the new dashboards :D

dstathis (approved):

PR #193, opened 2023-02-16
remove mention of LMA2

simskij:
Replaces a docstring mention of LMA2 with COS.

This PR does not warrant a new release of the lib and may be published the next time we make a change in need of publishing.


sed-i (approved):
> This PR does not warrant a new release of the lib

We should add NLP support to our CI :)

Abuelodelanada (approved):

Issue #196, opened 2023-02-24
Rename lokids and prometheusds to something more readable


aieri:
### Enhancement Proposal

The loki and prometheus data sources drop downs can be confusing for users (especially when you only have one data source). As a first easy enhancement they could be renamed to a more readable "Prometheus data source" / "Loki data source".

PR #197, opened 2023-03-01
Make the `All` option _actually_ pick all options

rbarry82:
## Issue
The "All" as empty string option isn't ok unless fuzzy searching is used, since `juju_application=""` won't match anything.

## Solution
Update the matchers (and injection) to use regexp searches. This is annoying in regexp searches due to the fact that we can't pass a template string to `cos-tool` (right now)

## Testing Instructions
Build a test Grafana with this library. Relate it to something (let's say Kafka). Make sure they are filled in.

## Release Notes
Make the `All` option _actually_ pick all options


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -256,7 +256,7 @@ def __init__(self, *args):
     {
         "allValue": None,
         "datasource": "${prometheusds}",
-        "definition": 'label_values(up{juju_model="$juju_model"},juju_model_uuid)',
+        "definition": 'label_values(up{juju_model=~"$juju_model"},juju_model_uuid)',
```
sed-i, line 259: So this is changed to `=~` because when "All" is selected then `$juju_model"` is empty string?

rbarry82: Yep.

> Custom all value
> This option is only visible if the Include All option is selected.
>
> Enter regex, globs, or lucene syntax in the Custom all value field to define the value of the All option.
>
> By default the All value includes all options in combined expression. This can become very long and can have performance problems. Sometimes it can be better to specify a custom all value, like a wildcard regex.
>
> In order to have custom regex, globs, or lucene syntax in the Custom all value option, it is never escaped so you will have to think about what is a valid value for your data source.

The difficulty with it is, to a degree, that some of the variables themselves (by using `label_values`) _also_ depend on this, and... their docs kind of say:

> Note: The Custom all value option on the variable must be blank for Grafana to format all values into a single string. If leave it blank, then the Grafana concatenates (adds together) all the values in the query. Something like value1,value2,value3. If a custom all value is used, then instead the value will be something like * or all.
>
> InfluxDB and Prometheus use regex expressions, so the same variable would be interpolated as (host1|host2|host3). Every value would also be regex escaped. If not, a value with a regex control character would break the regex expression.
>
> Automatic escaping and formatting can cause problems and it can be tricky to grasp the logic behind it. Especially for InfluxDB and Prometheus where the use of regex syntax requires that the variable is used in regex operator context.

So, assuming that even the baseline (`juju_model`) is set with the "default" `includeAll` value, where Grafana tries to do the "right thing", this will become `label_values(up{juju_model="(cos|other_model)"})`, and that percolates all the way down to everything else. These need to be regexp matchers for that kind of syntax to work.

lucabello (approved):

rbarry82:
> Resolves #196. (add it to your PR description so it closes the issue on merge!)
> 
> Would love for this to finally pass tests as well :D LGTM!

It _should_ pass now, rebased/merged on top of the working `main`. Only timing out because GH is GH...

PR #206, opened 2023-03-21
Handle empty keys on removal and in prometheus-ceph-exporter dashboards better

rbarry82:
Log an error if a key can't be found when removing a dashboard (if more than one exporter is related, the event.app.name will be the same, and the first relation-broken will purge it, as these seem to fire more than once in subordinates).

prometheus-ceph-exporter sends dashboards with None as explicit templating keys, which passes (theoretically) since the key exists, but None is not iterable. Add explicit None checks.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -219,7 +219,7 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 26
+LIBPATCH = 27
 
 logger = logging.getLogger(__name__)
 
@@ -1691,7 +1691,11 @@ class GrafanaDashboardAggregator(Object):
 
     def remove_dashboards(self, event: RelationBrokenEvent) -> None:
         """Remove a dashboard if the relation is broken."""
-        app_ids = _type_convert_stored(self._stored.id_mappings[event.app.name])  # type: ignore
+        app_ids = _type_convert_stored(self._stored.id_mappings.get(event.app.name, ""))  # type: ignore
+
+        if not app_ids:
+            logger.info("Could not look up stored dashboards for %s", event.app.name)  # type: ignore
+            return
 
         del self._stored.id_mappings[event.app.name]  # type: ignore
         for id in app_ids:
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1691,7 +1691,11 @@ def _update_remote_grafana(self, _: Optional[RelationEvent] = None) -> None:
 
     def remove_dashboards(self, event: RelationBrokenEvent) -> None:
         """Remove a dashboard if the relation is broken."""
-        app_ids = _type_convert_stored(self._stored.id_mappings[event.app.name])  # type: ignore
+        app_ids = _type_convert_stored(self._stored.id_mappings.get(event.app.name, ""))  # type: ignore
```
PietroPasotti, line 1694: instead of all those type:ignore, why not extract `app_name: str = event.app.name` and if necessary noqa that line alone?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1731,22 +1735,24 @@ def _strip_existing_datasources(self, dash: dict) -> dict:  # noqa: C901
...
-                        dash["templating"]["list"][i]["datasource"] = r"${prometheusds}"
-                    if (
-                        "name" in dash["templating"]["list"][i]
-                        and dash["templating"]["list"][i]["name"] == "host"
+                        and dash["templating"]["list"][i]["datasource"] is not None
                     ):
-                        dash["templating"]["list"][i] = REACTIVE_CONVERTER
+                        if "Juju" in dash["templating"]["list"][i].get("datasource", ""):
+                            dash["templating"]["list"][i]["datasource"] = r"${prometheusds}"
+                    # if (
```
PietroPasotti, line 1742: remove commented code?

PietroPasotti (approved):
Couple of minor nits, looks good for the rest.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
rbarry82: The `# type: ignore` is actually not around `event.app.name`, but around... basically every "magic" operation in `StoredState`, similar to all of the same fights you had when typing OF itself.

We just stuck on 3.5 for a long time for compat until OF moved up, and because we weren't sure what the long-term story would be around using libs on machine charms. It's on my near-term to-do list to move more of our stuff over to pyright, and since we broke the 3.5 leash, we can also add basic types around some of it and stop the `# type: ignore` (where almost every place touching `Stored*` needs it or mypy complains).

Point definitely well taken though ;)

PietroPasotti: Ahh right, my bad. I misinterpreted the code :)

PR #218, opened 2023-06-01
use pyright instead of mypy

lucabello:


PietroPasotti (approved):

PietroPasotti (changes requested):
there seem to be some failing itests. patching kubernetes_service isn't going well. Maybe some Path issues?

lucabello:
Fixed it!

sed-i (approved):

PietroPasotti (approved):
we're #ignoring and #Any'ing a bit too much for my taste, but fixing that (here) would be out of scope.

lucabello:
Indeed; though all the custom-event-stuff has to be ignored unfortunately because there's no typing before runtime; a solution would be to write the type structures separately just for the linter, but that indeed falls out of the scope of this :)

At least know the fact that something's potentially fishy is explicit though!

PR #226, opened 2023-07-05
Implement TLS

### LINKED ISSUE #225: Implement TLS and enable it when a cert is available
### Enhancement Proposal

Tracking issue: https://github.com/canonical/cos-lite-bundle/issues/75.

lucabello:
## Issue
Closes #225.

### DIFF
--- a/lib/charms/observability_libs/v0/cert_handler.py
+++ b/lib/charms/observability_libs/v0/cert_handler.py
@@ -1,0 +1,371 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""## Overview.
+
+This document explains how to use the `CertHandler` class to
+create and manage TLS certificates through the `tls_certificates` interface.
+
+The goal of the CertHandler is to provide a wrapper to the `tls_certificates`
+library functions to make the charm integration smoother.
+
+## Library Usage
+
+This library should be used to create a `CertHandler` object, as per the
+following example:
+
+```python
+self.cert_handler = CertHandler(
+    charm=self,
+    key="my-app-cert-manager",
+    peer_relation_name="replicas",
+    cert_subject="unit_name",  # Optional
+)
+```
+
+You can then observe the library's custom event and make use of the key and cert:
+```python
+self.framework.observe(self.cert_handler.on.cert_changed, self._on_server_cert_changed)
+
+container.push(keypath, self.cert_handler.key)
+container.push(certpath, self.cert_handler.cert)
+```
+
+This library requires a peer relation to be declared in the requirer's metadata. Peer relation data
+is used for "persistent storage" of the private key and certs.
+"""
+import json
+import socket
+from typing import List, Optional, Union
+
+try:
+    from charms.tls_certificates_interface.v2.tls_certificates import (  # type: ignore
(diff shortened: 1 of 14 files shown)


lucabello:
This PR still needs for grafana to trust the CA of its own certificate so it can communicate with the other bundle components.

metadata.yaml
```diff
@@ -57,6 +61,8 @@ provides:
 peers:
   grafana:
     interface: grafana_peers
+  replicas:
+    interface: grafana_replicas
```
PietroPasotti, line 65: we have two peer relations now, how about a note on what each one is used for?

src/charm.py
```diff
@@ -674,7 +691,16 @@ def _generate_grafana_config(self) -> str:
         can be set in ENV variables, but leave for expansion later so we can
         hide auth secrets
         """
-        return self._generate_database_config() if self.has_db else ""
+        db_config = self._generate_database_config() if self.has_db else ""
```
PietroPasotti, line 694: this reads a bit odd.
I'd refactor to use some if/elses:
```suggestion
cfgs = []
if self.has_db:
   cfgs.append(self._generate_db_config())
if self.has_cert:
    cfgs.append(self._generate_network_config())
return "\n".join(cfgs)
```

src/charm.py
```diff
@@ -1087,10 +1139,11 @@ def external_url(self) -> str:
             return web_external_url
 
         baseurl = ""
+        scheme = "https" if hasattr(self, "cert_handler") and self.cert_handler.cert else "http"
```
PietroPasotti, line 1142: I'd extract this to a `self.has_tls` or something like that.

why are we doing the hasattr call by the way? init ordering issues?

PietroPasotti (approved):
couple of nits, doesn't need a second round from me

src/charm.py
lucabello: I didn't extract it to a separate method since this is not necessarly `has_tls`; this is necessary to break a circular check as you guessed;

To instantiate `cert_handler`, we need the `external_url` property (to pass as _extra_sans_dns_); however, the `external_url` needs to check `cert_handler` to decide whether to use **https** or not.  
At the time of the call, `self.cert_handler = ...` has not been executed yet, thus the guard you're seeing here; it doesn't seem right to extract this one-liner to a method considering the scope is pretty narrow and it's not re-used anywhere else

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -0,0 +1,395 @@
...
+                sans_dns=self.sans_dns,
+            )
+
+            if renew and self._csr:
+                self.certificates.request_certificate_renewal(
+                    old_certificate_signing_request=self._csr.encode(),
+                    new_certificate_signing_request=csr,
+                )
+            else:
+                logger.info("Creating CSR for %s with DNS %s", self.cert_subject, self.sans_dns)
```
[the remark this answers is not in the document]
simskij: This is a false positive. We only log the subject and the SANs

metadata.yaml
sed-i: Why do we need two?

PR #231, opened 2023-07-23
[grafana_source] Vectorize refresh_event

### LINKED ISSUE #228: GrafanaSourceProvider: `refresh_event` should take a list
### Enhancement Proposal

Currently, the refresh_event arg is a scalar:

https://github.com/canonical/grafana-k8s-operator/blob/82377c9672f91f077fea7c7916ed79b2cb3d4b3b/lib/charms/grafana_k8s/v0/grafana_source.py#L324

It needs to be a vector.

sed-i:
## Issue
Need to be able to pass both ingress and tls events as refresh events.

## Solution
Vectorize the refresh_event arg.

Fixes #228.

## Context
https://github.com/canonical/prometheus-k8s-operator/pull/496

## Testing Instructions
NTA.

## Release Notes
Vectorize refresh_event for grafana_source provider.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -160,7 +160,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -405,8 +409,8 @@ def __init__(
         self._source_url = source_url
 
         self.framework.observe(events.relation_joined, self._set_sources_from_event)
-        if refresh_event:
-            self.framework.observe(refresh_event, self._set_unit_details)
+        for ev in refresh_event:
+            self.framework.observe(ev, self._set_unit_details)
```
Abuelodelanada, line 413: We should validate that every `ev` is a `BoundEvent` instance.

sed-i: I imagine the outcome of us checking and ops complaining would be the same: exception at utest time.

PR #238, opened 2023-08-02
Include dashboard url in get-admin-password action response

### LINKED ISSUE #92: Suggested feature: Add convenience action for providing all login information
This is purely a suggestion for a convenience function.

The `get-admin-password` action is great for getting the grafana dashboard password, but whenever I use it I'm really looking for:
* password
* url
* login (I realise it is fixed, but for completeness...)

There are ways of getting the other info programatically (could use `juju` to get the unit IP), but I think in most cases when someone wants password they probably also want to know the dashboard url to use the password on.  Would be great if they were all returned together (either in `get-admin-password`, or in a similar `get-dashboard-login-details` or whatever)

hemanthnakkina:
Add grafana dashboard url to response of action get-admin-password. 

Closes #92

## Issue

Currently the action get-admin-password returns only admin password but not the dashboard url.

## Solution

PR updates the response of action get-admin-password to include grafana dashboard url.

## Release Notes

Include grafana dashboard url in get-admin-password action response

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,5 +1,5 @@
 get-admin-password:
   description: |
-    Get the initial admin password for the Grafana web interface. This is generated at charm
-    deployment time. If the password has been changed, a notice of that fact will be returned by
-    this action instead.
+    Get the dashboard url and initial admin password for the Grafana web interface. Initial
+    admin password is generated at charm deployment time. If the password has been changed,
+    a notice of that fact will be returned by this action instead.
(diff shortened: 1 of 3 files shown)


simskij (dismissed):
Thank you for your contribution! However, I would much rather have the existing action enhanced to include the needed information than another one added.

hemanthnakkina:
> Thank you for your contribution! However, I would much rather have the existing action enhanced to include the needed information than another one added.

Thanks for the review. Updated the existing action get-admin-password instead of introducing new action.

sed-i (approved):
Thanks @hemanthnakkina!

lucabello (approved):

PR #251, opened 2023-08-24
remove insecure-skip-verify for HTTPS datasources

lucabello:
## Issue
When the charm doesn't have a relation to a CA, we currently use "tlsSkipVerify" to still use HTTPS datasources. We want to remove this behavior and only allow HTTPS datasources to work over TLS, meaning the Grafana charm will need to trust the datasource's CA (either by being related to the same CA charm, or eventually getting it by a future `certificate-transfer` interface).

## Solution
Remove the piece of code adding `tlsSkipVerify`.


simskij (approved):

simskij:
Once the integration test passes, I'm in favor of merging this.

PR #257, opened 2023-08-30
Fix `stripPrefix`

sed-i:
## Issue
The fix in #253 removed `stripPrefix` as a workaround, but we still want the stripPrefix functionality.

Depends on:
- https://github.com/canonical/prometheus-k8s-operator/pull/520

## Solution
Revert #253 and take an approach similar to https://github.com/canonical/prometheus-k8s-operator/pull/514.

Refs:
- https://github.com/grafana/grafana/blob/main/conf/defaults.ini
- https://grafana.com/tutorials/run-grafana-behind-a-proxy/

In tandem with:
- https://github.com/canonical/traefik-route-k8s-operator/pull/41
- https://github.com/canonical/traefik-k8s-operator/pull/234

Drive-by fixes:
- Differentiate scheme between grafana and traefik: In tls termination, traefik's scheme is https but grafana's is http.
- Use envvars instead of config ini.
- fetch-lib

## Testing Instructions
Navigate with a browser to the grafana endpoint using each of the newly created bundle files.

## Release Notes
Fix stripPrefix.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -219,7 +219,7 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 35
+LIBPATCH = 34
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -219,7 +219,7 @@ def __init__(self, *args):
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 34
+LIBPATCH = 35
```
sed-i, line 222: `self._update_all_dashboards_from_dir()` was previously added at the same time another PR merged, both bumped to 34.
This change brings it back in sync.

PR #278, opened 2023-11-24
IAM-224-Oauth interface implementation

bencekov:
## Issue

Grafana is Oauth 2.0 capable, but the charm currently can't use the Identity Platform's oauth interface.

## Solution

This pr implements integration with the oauth interface, and includes unit and integration tests to verify correct behavior.

## Context

Relevant projects are:
playwright
[Identity bundle](https://github.com/canonical/iam-bundle)
[Hydra charm](https://github.com/canonical/hydra-operator)

## Testing Instructions

Use tox for both unit and integration test.


src/charm.py
```diff
@@ -1303,6 +1347,33 @@ def _update_cert(self):
...
+        return ClientConfig(
+            os.path.join(self.external_url, "login/generic_oauth"),
+            OAUTH_SCOPES,
+            OAUTH_GRANT_TYPES,
+        )
+
+    def _on_oauth_info_changed(self, event: OAuthInfoChangedEvent) -> None:
+        """Event handler for the oauth_info_changed event."""
+        if not self.unit.is_leader():
+            return
```
nsklikas, line 1361: This guard is not needed, all the grafana units should update their config with the new client credentials

src/charm.py
```diff
@@ -1303,6 +1347,33 @@ def _update_cert(self):
...
+
+        self.oauth.update_client_config(client_config=self._oauth_client_config)
```
nsklikas, line 1363: The URL of the application has not changed, no reason to update the databag.

src/charm.py
```diff
@@ -1303,6 +1347,33 @@ def _update_cert(self):
...
+        logger.info(f"Received oauth provider info: {self.oauth.get_provider_info()}")
+
+        if not event.client_id or not event.client_secret_id:
+            return
```
nsklikas, line 1367: These should always be there, am I missing something?

src/charm.py
```diff
@@ -262,11 +283,16 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         Args:
             event: a :class:`ConfigChangedEvent` to signal that something happened
         """
+        if self.model.relations[OAUTH]:
```
nsklikas, line 286: This is checked here, but not on https://github.com/canonical/grafana-k8s-operator/pull/278/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR294. Which one is correct?

Perhaps we should add a check in the `oauth` relation and remove this from the charm code?

src/charm.py
```diff
@@ -97,6 +103,10 @@
 PEER = "grafana"
 PORT = 3000
 
+OAUTH = "oauth"
+OAUTH_SCOPES = "openid email"
+OAUTH_GRANT_TYPES = ["authorization_code", "refresh_token"]
```
nsklikas, line 108: Do we want to use refresh tokens? If yes then grafana requires some extra config, see https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#configure-a-refresh-token

tests/integration/test_grafana_oauth.py
```diff
@@ -0,0 +1,112 @@
...
+
+async def test_oauth_login_with_identity_bundle(
+    ops_test: OpsTest, page: Page, context: BrowserContext
+) -> None:
+    external_idp_manager = ExternalIdpManager(ops_test=ops_test)
+
+    grafana_proxy = await get_reverse_proxy_app_url(
+        ops_test, OAUTH_RELATION.OAUTH_PROXY, "grafana"
+    )
+    redirect_login = f"{grafana_proxy}login"
```
nsklikas, line 83: Let's use some `join` function for this (e.g. like we do on https://github.com/canonical/grafana-k8s-operator/pull/278/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR1353).

tests/integration/oauth_tools/dex.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def update_redirect_uri(self, redirect_uri: str) -> None:
+        # Updates the redirect uri configuration for the identity provider.
+        if not redirect_uri:
+            logger.info("Empty parameter for redirect_uri")
+            return
+        self._redirect_uri = redirect_uri
+        self._apply_dex_resources()
+
+    def close(self) -> None:
```
nsklikas, line 53: I think `remove` or `delete` would be a more appropriate name.

tests/integration/test_grafana_oauth.py
```diff
@@ -0,0 +1,112 @@
...
+    redirect_url = grafana_proxy + "?*"
+    await verify_page_loads(page=page, url=redirect_url)
+
+    # Verifying that the login flow was successful is application specific.
+    # The test uses Grafana's /api/user endpoint to verify the session cookie is valid
+    grafana_session_cookie = await get_cookie_from_browser_by_name(
+        browser_context=context, name="grafana_session"
+    )
+    request = requests.get(
+        f"{grafana_proxy}api/user",
```
nsklikas, line 106: Let's use some join function for this (e.g. like we do on https://github.com/canonical/grafana-k8s-operator/pull/278/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR1353).

tests/integration/test_grafana_oauth.py
```diff
@@ -0,0 +1,112 @@
...
+        headers={"Cookie": f"grafana_session={grafana_session_cookie}"},
+        verify=False,
+    )
+    assert request.status_code == 200
```
nsklikas, line 110: Should we validate something about the user here? Do we get back any of the user's attributes? (e.g. email)

tests/integration/oauth_tools/oauth_test_helper.py
```diff
@@ -0,0 +1,143 @@
...
+        await expect(page).to_have_url(re.compile(rf"{redirect_login_url}*"))
+
+
+async def click_on_sign_in_button_by_text(page: Page, text: str):
+    """Convenience function for retrieving the Oauth Sign In button by its displayed text."""
+    async with page.expect_navigation():
+        await page.get_by_text(text).click()
+
+
+async def click_on_sign_in_button_by_alt_text(page: Page, alt_text: str):
```
nsklikas, line 95: This is not used.

tests/integration/oauth_tools/oauth_test_helper.py
```diff
@@ -0,0 +1,143 @@
...
+    await page.get_by_placeholder("email address").click()
+    await page.get_by_placeholder("email address").fill(EXTERNAL_USER_EMAIL)
+    await page.get_by_placeholder("password").click()
+    await page.get_by_placeholder("password").fill(EXTERNAL_USER_PASSWORD)
+    await page.get_by_role("button", name="Login").click()
+
+
+async def verify_page_loads(page: Page, url: str):
+    """Convenience function for verifying that the correct url has been loaded."""
+    await page.wait_for_url(url)
```
nsklikas, line 125: Do we need to wrap around a single function call?

nsklikas (commented):
I think that the `oauth_tools.constants.OAUTH_RELATION` var should be removed. It is used in too many places and I find it confusing.

Also not sure if all of the helper functions make much sense, e.g. I would expect `access_application_login_page` to simply navigate me to a page, but it also expects a redirect to happen. This seems to specific to this use case and I am not sure if we need to abstract it.

src/charm.py
bencekov: I checked, there's a guard if statement for these before firing the event in the charm lib.

nsklikas: Let's remove this check then

src/charm.py
nsklikas: IMHO The charm should have to make as little changes as possible, if a guard like that is needed then I think we need to add a `ready` function (or something like that) in the library itself.

tests/integration/test_grafana_oauth.py
bencekov: added check for email

src/charm.py
bencekov: ok

tests/integration/oauth_tools/dex.py
bencekov: changed name to remove_idp_service

src/charm.py
bencekov: I made all the listed configurations necessary to enable using refresh tokens. However I'm not sure how to verify that it worked.

bencekov: I verified that refresh tokens work.

tests/integration/test_grafana_oauth.py
```diff
@@ -0,0 +1,124 @@
...
+        browser_context=context, name="grafana_session"
+    )
+    request = requests.get(
+        os.path.join(grafana_proxy, "api/user"),
+        headers={"Cookie": f"grafana_session={grafana_session_cookie}"},
+        verify=False,
+    )
+    assert request.status_code == 200
+    assert request.status_code == 200
+    request.raise_for_status()
```
nsklikas, line 121: You already verified the status is 200, this will never raise
```suggestion
```

.github/workflows/pull-request.yaml
```diff
@@ -10,3 +10,6 @@ jobs:
     name: PR
     uses: canonical/observability/.github/workflows/charm-pull-request.yaml@main
     secrets: inherit
+    with:
+      ip-range-start: 10.64.140.43
+      ip-range-end: 10.64.140.49
```
nsklikas, line 15: We only need 2 ips, right? Let's request for the minimum amount of ips needed

.github/workflows/release.yaml
```diff
@@ -9,3 +9,6 @@ jobs:
   release:
     uses: canonical/observability/.github/workflows/charm-release.yaml@main
     secrets: inherit
+    with:
+      ip-range-start: 10.64.140.43
+      ip-range-end: 10.64.140.49
```
nsklikas, line 14: We only need 2 ips, right? Let's request for the minimum amount of ips needed

src/charm.py
```diff
@@ -709,11 +732,17 @@ def _generate_grafana_config(self) -> str:
...
         hide auth secrets
+
+        The feature toggle accessTokenExpirationCheck is also set here. It's needed
+        for the oauth relation to provide refresh tokens.
         """
         configs = []
         if self.has_db:
             configs.append(self._generate_database_config())
 
+        if self.model.relations[OAUTH]:
```
nsklikas, line 743: Let's keep it consistent
```suggestion
        if self.oauth.is_client_created():
```

src/charm.py
```diff
@@ -285,11 +304,15 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         Args:
             event: a :class:`ConfigChangedEvent` to signal that something happened
         """
+        self.oauth.update_client_config(client_config=self._oauth_client_config)
```
nsklikas, line 307: Is this actually needed here? This needs to run only if the ingress URL changed.

tests/unit/test_charm.py
```diff
@@ -372,6 +377,77 @@ def test_config_is_updated_with_authentication_config(self):
...
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_CLIENT_ID"], "grafana_client_id"
+        )
+        self.assertEqual(services["environment"]["GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET"], "s3cR#T")
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_SCOPES"], "openid email offline_access"
+        )
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_AUTH_URL"],
+            "https://example.oidc.com/oauth2/auth",
```
nsklikas, line 433:
```suggestion
                    oauth_provider_info["authorization_endpoint"],
```

tests/unit/test_charm.py
```diff
@@ -372,6 +377,77 @@ def test_config_is_updated_with_authentication_config(self):
...
+        )
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_TOKEN_URL"],
+            "https://example.oidc.com/oauth2/token",
```
nsklikas, line 437:
```suggestion
            oauth_provider_info["token_endpoint"]
```

tests/unit/test_charm.py
```diff
@@ -372,6 +377,77 @@ def test_config_is_updated_with_authentication_config(self):
...
+        )
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_API_URL"],
+            "https://example.oidc.com/userinfo",
```
nsklikas, line 441:
```suggestion
            oauth_provider_info["userinfo_endpoint"]
```

tests/unit/test_charm.py
```diff
@@ -372,6 +377,77 @@ def test_config_is_updated_with_authentication_config(self):
...
+        )
+
+        self.assertEqual(services["environment"]["GF_AUTH_GENERIC_OAUTH_ENABLED"], "True")
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_NAME"], "external identity provider"
+        )
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_CLIENT_ID"], "grafana_client_id"
+        )
+        self.assertEqual(services["environment"]["GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET"], "s3cR#T")
```
nsklikas, line 427: Minor comment, but let's put these in a variable

tests/integration/oauth_tools/constants.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+OAUTH_RELATION = collections.namedtuple(
+    "OAUTH_RELATION", ["OAUTH_APPLICATION", "OAUTH_INTERFACE", "OAUTH_PROXY", "OAUTH_CERTIFICATES"]
+)(
+    OAUTH_APPLICATION="hydra",
+    OAUTH_INTERFACE="oauth",
+    OAUTH_PROXY="traefik-public",
+    OAUTH_CERTIFICATES="self-signed-certificates",
+)
```
nsklikas, line 45: I don't really like this. No need to abstract `Hydra` or `Traefik`. IMHO it makes the code easier to read when you know which apps you are integrating.

I would suggest to remove the whole `OAUTH_RELATION` var

tests/integration/test_grafana_oauth.py
bencekov: this was a rebase error. sorry about this.

tests/integration/oauth_tools/constants.py
bencekov: I see your point. fixed.

nsklikas:
With the update of `python-libjuju` I had to change some other tests due to changes in `wait_for_idle`. Everything else seems to be working fine, should be ready for review.

nsklikas:
@sed-i @PietroPasotti @simskij this PR is ready for review. It's gotten a little big, let us know if we can help you reviewing it.

If you want a visual representation of the browser test, you can run:
```console
. .tox/integration/bin/activate
PYTHONPATH=. PWDEBUG=1 pytest  -k test_grafana_oauth tests/integration/
```

src/charm.py
```diff
@@ -1373,9 +1419,37 @@ def _on_trusted_certificate_removed(self, event: CertificateRemovedEvent):
...
+    def _oauth_client_config(self) -> ClientConfig:
+        return ClientConfig(
+            os.path.join(self.external_url, "login/generic_oauth"),
+            OAUTH_SCOPES,
+            OAUTH_GRANT_TYPES,
+        )
+
+    def _on_oauth_info_changed(self, event: OAuthInfoChangedEvent) -> None:
+        """Event handler for the oauth_info_changed event."""
+        logger.info(f"Received oauth provider info: {self.oauth.get_provider_info()}")
```
sed-i, line 1434: Debug leftover?

src/charm.py
```diff
@@ -1373,9 +1419,37 @@ def _on_trusted_certificate_removed(self, event: CertificateRemovedEvent):
...
+
+        self._configure()
+
+    def _on_oauth_info_removed(self, event: OAuthInfoRemovedEvent) -> None:
+        """Event handler for the oauth_info_removed event."""
+        logger.info("Oauth relation is broken, removing related settings")
```
sed-i, line 1440: Debug leftover?

tests/unit/test_charm.py
```diff
@@ -75,6 +75,14 @@
 
 """
 
+
```
sed-i, line 78: Should we have a separate file, e.g. `test_oauth`?

tests/unit/test_charm.py
```diff
@@ -372,6 +380,79 @@ def test_config_is_updated_with_authentication_config(self):
...
+        # assert that generic_oauth config is updated
+        services = (
+            self.harness.charm.containers["workload"].get_plan().services["grafana"].to_dict()
+        )
+
+        self.assertEqual(services["environment"]["GF_AUTH_GENERIC_OAUTH_ENABLED"], "True")
+        self.assertEqual(
+            services["environment"]["GF_AUTH_GENERIC_OAUTH_NAME"], "external identity provider"
+        )
+        self.assertEqual(
```
sed-i, line 427: Are all the envvars below strictly necessary for a valid oauth config? Asking to reduce brittleness of the test.

It could be valuable to also assert:
- without an oauth relation, or with empty oauth reldata, grafana config is unaffected.
- after removing the relation, grafana config is clear of oauth.
- simulated upgrade: mock the value of e.g. SCOPES and confirm it propagated to the config after emitting the upgrade sequence.

tests/unit/test_charm.py
```diff
@@ -372,6 +380,79 @@ def test_config_is_updated_with_authentication_config(self):
...
+                "client_secret_id": secret_id,
+            },
+        )
+
+        # assert that generic_oauth config is updated
+        services = (
+            self.harness.charm.containers["workload"].get_plan().services["grafana"].to_dict()
+        )
+
+        self.assertEqual(services["environment"]["GF_AUTH_GENERIC_OAUTH_ENABLED"], "True")
```
sed-i, line 423: Using `env = services["environment"]` would simplify the statements below.

sed-i (commented):
Nice!
- First-pass comments are in.
- Rebase on top of main?

src/charm.py
```diff
@@ -290,6 +309,8 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
 
     def _on_ingress_ready(self, _) -> None:
         """Once Traefik tells us our external URL, make sure we reconfigure Grafana."""
+        self.oauth.update_client_config(client_config=self._oauth_client_config)
```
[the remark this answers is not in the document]
nsklikas: My understanding is that this was done here because the client config needs to change only when the grafana ingress URL changes. Of course, given juju's wild and rebellious behavior (ie data is still there when you expect it to be gone), maybe it would be simpler it call it in `_configure` since it is not a very expensive call. 

I will move it since you are fine with it.

src/charm.py
```diff
@@ -709,11 +730,17 @@ def _generate_grafana_config(self) -> str:
...
         hide auth secrets
+
+        The feature toggle accessTokenExpirationCheck is also set here. It's needed
+        for the oauth relation to provide refresh tokens.
         """
         configs = []
         if self.has_db:
             configs.append(self._generate_database_config())
 
+        if self.oauth.is_client_created():
```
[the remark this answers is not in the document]
nsklikas: My understanding is that on relation broken the behavior is unpredictable, the data may be there or it may not be there. Is there a consistent, clean way to handle this? Only thing that I can think of is adding a flag (or something along those lines) that will be set when `_on_oauth_info_removed` calls `_configure` that will signify that the oauth relation config should be removed.

We can add something like this if you want, but IMO it will make the code dirtier-harder to maintain so I am not sure if is worth it.

src/charm.py
```diff
@@ -908,6 +935,25 @@ def _build_layer(self) -> Layer:
...
             )
 
+        if self.oauth.is_client_created():
+            oauth_provider_info = self.oauth.get_provider_info()
+
+            extra_info.update(
+                {
+                    "GF_AUTH_GENERIC_OAUTH_ENABLED": "True",
+                    "GF_AUTH_GENERIC_OAUTH_NAME": "external identity provider",
+                    "GF_AUTH_GENERIC_OAUTH_CLIENT_ID": cast(str, oauth_provider_info.client_id),
```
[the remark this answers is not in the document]
nsklikas: Not sure if I understand this comment, you mean to remove the call to `get_provider_info` and make `client_id` a property?

tests/unit/test_charm.py
nsklikas: This is up to you (you will have to maintain this code). IMO, given how the grafana tests are structured, it would make sense to move all the tests related to the oauth relation to a separate file.

src/charm.py
```diff
@@ -908,6 +931,27 @@ def _build_layer(self) -> Layer:
...
+                    "GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET": cast(
+                        str, oauth_provider_info.client_secret
+                    ),
+                    "GF_AUTH_GENERIC_OAUTH_SCOPES": OAUTH_SCOPES,
+                    "GF_AUTH_GENERIC_OAUTH_AUTH_URL": oauth_provider_info.authorization_endpoint,
+                    "GF_AUTH_GENERIC_OAUTH_TOKEN_URL": oauth_provider_info.token_endpoint,
+                    "GF_AUTH_GENERIC_OAUTH_API_URL": oauth_provider_info.userinfo_endpoint,
+                    "GF_AUTH_GENERIC_OAUTH_USE_REFRESH_TOKEN": "True",
+                    # This toggle will be removed on grafana v10.3
+                    "GF_FEATURE_TOGGLES_ENABLE": "accessTokenExpirationCheck",
```
nsklikas, line 951: Should we add a TODO to remove this? (relevant comment https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v10-2/#configure-refresh-token-handling-separately-for-oauth-providers)

tests/unit/test_charm.py
nsklikas: > after removing the relation, grafana config is clear of oauth.

I guess this explains your comment about `is_client_created` not handling a broken relation. We can't test this using harness, harness throws an error because the lib is trying to access a databag that is going away. AFAICT this is not the juju behavior, which is weird.

> simulated upgrade: mock the value of e.g. SCOPES and confirm it propagated to the config after emitting the upgrade sequence.

Are there other tests like this which I can copy?

src/charm.py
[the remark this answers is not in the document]
sed-i: [ops/940](https://github.com/canonical/operator/issues/940) is underway but seems like it would still be an opt-in, i.e. we would need to condition on the new `Relation.broken` (or `Relation.active`).

But: perhaps we can just always render that part of the config, without checking for any relations at all?

src/charm.py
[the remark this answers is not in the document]
sed-i: Correct.

tests/unit/test_charm.py
sed-i: Agreed. This would also make `setUp()` work better for you.

tests/unit/test_charm.py
sed-i: > We can't test this using harness, harness throws an error

This is odd indeed. Submit an issue? In any case, here's a somewhat relevant example:
```
grafana-k8s-operator/tests/unit/test_external_url.py
135-        # WHEN the traefik relation is removed
136-        with patch.object(TraefikRouteRequirer, "external_host", new=""):
137:            self.harness.remove_relation_unit(rel_id, "traefik-app/0")
138:            self.harness.remove_relation(rel_id)
139-
140-            # THEN root url and subpath envs are undefined (because fqdn is a bare hostname)
141-            self.assertEqual(self.get_pebble_env()["GF_SERVER_SERVE_FROM_SUB_PATH"], "True")
142-            self.assertEqual(
143-                self.get_pebble_env()["GF_SERVER_ROOT_URL"],
144-                "http://grafana-k8s-0.testmodel.svc.cluster.local:3000",
145-            )
146-            self.assertTrue(self.is_service_running())
```

> Are there other tests like this which I can copy?

Not that I know of. Testing upgrade is [tricky](https://github.com/canonical/operator/issues/879) though. Perhaps it's better to not do it right now.

src/charm.py
sed-i: Good question. Generally speaking, we would need layer config to depend on workload version.
Workload version should be usually [obtainable](https://github.com/canonical/alertmanager-k8s-operator/blob/950777a70e17e5f6a20eea893b2d8daebc5f34ad/src/alertmanager.py#L141) if we are about to update the layer.

I think a plain TODO will do, as I imagine having `"accessTokenExpirationCheck"` with v10.3 won't cause grafana to crash.

But wdyt @simskij @Abuelodelanada @lucabello @mmkay ?

mmkay: A plain TODO sounds fine for me. If we _really_ want to be on the safe side, we could add a 10.3 upgrade issue in the repo mentioning the removal of this toggle.

src/charm.py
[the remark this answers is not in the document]
nsklikas: I think that I'd prefer to keep it this way tbh, `get_provider_info` returns all the config needed for most apps to configure their oauth, IMO it is clearer and easier to use rather than having 5 properties that the app will have to check on.

src/charm.py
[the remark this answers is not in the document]
nsklikas: I removed that part of the config altogether and switched it to an env var.

tests/unit/test_charm.py
nsklikas: I added a test that xfails for relation broken. If it's ok we can open an issue on this repo to keep track of it. So that, once https://github.com/canonical/operator/issues/940 is fixed, we can enable the test.

sed-i: > harness does not allow you to access the databag of a breaking relation

Do we need to access it at all? Aren't we interested in the resulting grafana config, e.g. what envvars are there?

sed-i (commented):
Looking good!
I'd like us to iron out two related items:
- Polish integration tests. The volume of newly added files seems unreasonable to me.
- A bundle for manual testing.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pyyaml
 urllib3
 jsonschema
 cryptography
+git+https://github.com/canonical/iam-bundle@IAM-640-outsource-oauth-tools-package#egg=oauth_tools
```
shipperizer, line 9: lets stick with `git+ssh`

i think issue with the CI is the fact that in the tox.ini we don't install deps from `requirements.txt`

simskij: Sounds reasonable. @bencekov, @nsklikas, @natalian98, is this PR making https://github.com/canonical/grafana-k8s-operator/pull/213 obsolete?

nsklikas: Yes, you can close that

src/charm.py
```diff
@@ -256,6 +264,16 @@ def __init__(self, *args):
...
         )
+        # oauth relation
+        self.oauth = OAuthRequirer(self, self._oauth_client_config)
+
+        # oauth relation observations
+        self.framework.observe(
+            self.oauth.on.oauth_info_changed, self._on_oauth_info_changed  # pyright: ignore
+        )
+        self.framework.observe(
+            self.oauth.on.oauth_info_removed, self._on_oauth_info_changed  # pyright: ignore
```
PietroPasotti, line 275: it's best practice to have dedicated event handlers for each event type.
so I'd recommend we:
- add an `_on_oauth_info_removed` method

if that means repeating a lot of code, lift the shared code to a third method which both handlers delegate to

src/charm.py
```diff
@@ -814,6 +834,10 @@ def restart_grafana(self) -> None:
             # now that this is done, build_layer should include cert and key into the config and we'll
             # be sure that the files are actually there before grafana is (re)started.
 
+        # If available, we collect all trusted certs from the receive-ca-cert relation
+        # we do this here, downstream from a container readiness check
+        self._update_trusted_ca_certs()
+
```
PietroPasotti, line 840: why is this change part of this PR?

PietroPasotti (commented):
couple of minor recommendations/clarification requests from my side, rest looks good.

src/charm.py
nsklikas: https://github.com/canonical/grafana-k8s-operator/pull/278#discussion_r1446751611

Not sure which approach we should follow

sed-i: @PietroPasotti we had opposite suggestions :)
I may have forgotten a best practice we established, but resolving for now.

sed-i (approved):

lucabello (approved):

Issue #277, opened 2023-11-23
Implement integration certificate_transfer with self-signed-certificates-operator


bencekov:
### Enhancement Proposal

We need this enhancement for Grafana to be able to integrate with the Identity Platform.

Grafana needs to trust the tls certificate of the proxy service (traefik) we use to complete an Oauth 2.0 login flow with the Identity Platform. Our proxy gets it's certificate from the [self-signed-certificates-operator](https://github.com/canonical/self-signed-certificates-operator). For authentication to work, the grafana charm needs to implement receiving the trusted cert from the self-signed-certificates-operator charm's certificate_transfer integration.

sed-i:
Reference impl for adding cert transfer:
- https://github.com/canonical/traefik-k8s-operator/pull/204
- https://github.com/canonical/traefik-k8s-operator/pull/224

Could be a good opportunity to introduce the "tls truststore".

lucabello:
We're planning the work for this feature; in the meantime, one possible workaround could be relating Grafana to the same **self-signed-certificates-operator** as Traefik; this will make Grafana trust the same CA and things should work.

sed-i:
Impl'd in #279. Closing.

PR #279, opened 2023-11-29
implement certificate transfer interface

bencekov:
## Issue

This PR is a solution for issue #277. 

## Solution

This pr implements handling the self-signed-certificate's certificate_transfer interface.
When the certificate_transfer integration databag is updated with a new certificate the following steps are taken:
- Grafana app restarts.
- The cert from the relation is stored in a file in directory /usr/local/share/ca-certificates.
- Trusted certificates are updated.

When the cert is revoked by the relation, the file in /usr/local/share/ca-certificates is deleted, grafana is restarted,
and the trusted certificates are updated.

## Testing Instructions

An integration test is included to check for the received certificate being included among the trusted certificates.
```tox -e integration```

### DIFF
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,1 +1,0 @@
-lib/charms/*
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,0 +1,1 @@
+lib/charms/*
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -52,6 +52,13 @@ requires:
     interface: tls-certificates
     limit: 1
     description: Certificate and key files for Grafana to use with TLS.
+  receive-ca-cert:
+    interface: certificate_transfer
+    description: |
+      Receive a CA cert for grafana to trust.
+      This relation can be used with a local CA to obtain the CA cert that was used to sign proxied
+      endpoints.
+    limit: 1
 
 provides:
   metrics-endpoint:
@@ -69,7 +76,7 @@ resources:
   grafana-image:
     type: oci-image
     description: upstream docker image for Grafana
-    upstream-source: docker.io/ubuntu/grafana:10-22.04
+    upstream-source: ghcr.io/canonical/grafana:dev
   litestream-image:
     type: oci-image
     description: upstream image for sqlite streaming
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -77,6 +77,8 @@ resources:
     type: oci-image
     description: upstream docker image for Grafana
     upstream-source: ghcr.io/canonical/grafana:dev
+    # TODO revert to dockerhub once the oci-factory CI is fixed
+    # upstream-source: docker.io/ubuntu/grafana:10-22.04
   litestream-image:
     type: oci-image
     description: upstream image for sqlite streaming
--- a/src/charm.py
+++ b/src/charm.py
@@ -49,6 +49,11 @@ from charms.observability_libs.v0.kubernetes_compute_resources_patch import (
     adjust_resource_requirements,
 )
 from charms.observability_libs.v0.cert_handler import CertHandler
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateAvailableEvent,
+    CertificateRemovedEvent,
+    CertificateTransferRequires,
+)
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
 from ops.charm import (
@@ -97,6 +102,11 @@ DATABASE = "database"
 PEER = "grafana"
 PORT = 3000
 
+# Template for storing trusted certificate in a file.
+TRUSTED_CA_TEMPLATE = string.Template(
+    "/usr/local/share/ca-certificates/trusted-ca-cert-$rel_id-ca.crt"
+)
+
 
 class GrafanaCharm(CharmBase):
     """Charm to run Grafana on Kubernetes.
@@ -130,6 +140,9 @@ class GrafanaCharm(CharmBase):
             peer_relation_name="replicas",
             extra_sans_dns=[socket.getfqdn()],
         )
+
+        # -- trusted_cert_transfer
+        self.trusted_cert_transfer = CertificateTransferRequires(self, "receive-ca-cert")
 
         # -- ingress via raw traefik_route
         # TraefikRouteRequirer expects an existing relation to be passed as part of the constructor,
@@ -228,6 +241,16 @@ class GrafanaCharm(CharmBase):
         # -- cert_handler observations
         self.framework.observe(
             self.cert_handler.on.cert_changed, self._on_server_cert_changed  # pyright: ignore
+        )
+
+        # -- trusted_cert_transfer observations
+        self.framework.observe(
+            self.trusted_cert_transfer.on.certificate_available,
+            self._on_trusted_certificate_available,  # pyright: ignore
+        )
+        self.framework.observe(
+            self.trusted_cert_transfer.on.certificate_removed,
+            self._on_trusted_certificate_removed,  # pyright: ignore
         )
 
         # self.catalog = CatalogueConsumer(charm=self, item=self._catalogue_item)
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: src/charm.py [hunk 1]
```diff
@@ -1300,8 +1335,45 @@ def _update_cert(self):
...
+        if not self.containers["workload"].can_connect():
+            logger.warning("Cannot connect to Pebble. Deferring event.")
+            event.defer()
+            return
+
+        self.restart_grafana()
+
+    def _update_trusted_ca_certs(self):
+        """This function receives the trusted certificates from the certificate_transfer integration."""
+        """Grafana needs to restart to use newly received certificates. Certificates attached to the
```
nsklikas, line 1350: Something went wrong here
```suggestion
        """This function receives the trusted certificates from the certificate_transfer integration.
        
        Grafana needs to restart to use newly received certificates. Certificates attached to the
```

src/charm.py
```diff
@@ -1300,8 +1335,45 @@ def _update_cert(self):
...
 
-        container.exec(["update-ca-certificates", "--fresh"]).wait()
-        subprocess.run(["update-ca-certificates", "--fresh"])
+        self._update_system_certs()
+
+    def _on_trusted_certificate_available(self, event: CertificateAvailableEvent):
+        if not self.containers["workload"].can_connect():
+            logger.warning("Cannot connect to Pebble. Deferring event.")
+            event.defer()
+            return
```
nsklikas, line 1344: I think that `restart_grafana` should make this check.

src/charm.py
```diff
@@ -1300,8 +1335,45 @@ def _update_cert(self):
...
+        for relation in self.model.relations.get(self.trusted_cert_transfer.relationship_name, []):
+            # For some reason, relation.units includes our unit and app. Need to exclude them.
+            for unit in set(relation.units).difference([self.app, self.unit]):
+                # Note: this nested loop handles the case of multi-unit CA, each unit providing
+                # a different ca cert, but that is not currently supported by the lib itself.
+                cert_path = TRUSTED_CA_TEMPLATE.substitute(rel_id=relation.id)
+                if cert := relation.data[unit].get("ca"):
+                    container.push(cert_path, cert, make_dirs=True)
+
+        self._update_system_certs()
```
nsklikas, line 1368: Why refresh the charm container's certs here?

src/charm.py
nsklikas: Fair

src/charm.py
```diff
@@ -1268,6 +1298,11 @@ def _on_server_cert_changed(self, _):
         self._update_cert()
         self._configure()
 
+    def _update_system_certs(self):
+        container = self.containers["workload"]
+        container.exec(["update-ca-certificates", "--fresh"]).wait()
+        subprocess.run(["update-ca-certificates", "--fresh"])
```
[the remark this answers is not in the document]
bencekov: It's how it's done for cert_handler gets reconfigured (the thing that makes https to grafana work). For our purpose it's probably unnecessary.

src/charm.py
bencekov: the function up until  line 1368 only copies the certificate to /usr/local/share/ca-certificates/trusted-ca-cert. It gets into /etc/ssl/certs/ca-certificates.crt, as well as into an internal database (I think) by running update-ca-certificate.

src/charm.py
```diff
@@ -1268,6 +1298,11 @@ def _on_server_cert_changed(self, _):
         self._update_cert()
         self._configure()
 
+    def _update_system_certs(self):
+        container = self.containers["workload"]
+        container.exec(["update-ca-certificates", "--fresh"]).wait()
```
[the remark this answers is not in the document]
bencekov: fixed

src/charm.py
bencekov: fixed

.wokeignore
sed-i, line 1: We should have a more specific wokeignore. This one includes libs owned by the charm.
See https://github.com/canonical/traefik-k8s-operator/blob/main/lib/.wokeignore

metadata.yaml
```diff
@@ -69,7 +76,7 @@ resources:
   grafana-image:
     type: oci-image
     description: upstream docker image for Grafana
-    upstream-source: docker.io/ubuntu/grafana:10-22.04
+    upstream-source: ghcr.io/canonical/grafana:dev
```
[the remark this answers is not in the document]
sed-i: Let's keep the old one as a comment?

```suggestion
    upstream-source: ghcr.io/canonical/grafana:dev
    # TODO revert to dockerhub once the oci-factory CI is fixed
    # upstream-source: docker.io/ubuntu/grafana:10-22.04
```

src/charm.py
sed-i: Hopefully with pebble notices we'd be able to get rid of the defer without refactoring the charm to common-exit-hook.

src/charm.py
```diff
@@ -1303,6 +1333,45 @@ def _update_cert(self):
...
+        Grafana needs to restart to use newly received certificates. Certificates attached to the
+        relation need to be pulled before Grafana is started.
+        This function is needed because relation events are not emitted on upgrade, and because we
+        do not have (nor do we want) persistent storage for certs.
+        """
+        if not self.model.get_relation(relation_name=self.trusted_cert_transfer.relationship_name):
+            return
+
+        container = self.containers["workload"]
+        for relation in self.model.relations.get(self.trusted_cert_transfer.relationship_name, []):
```
sed-i, line 1356: Do we need the `if not` above in addition to this `.get()`?

sed-i (approved):

src/charm.py
bencekov: I see your point. I would leave it in so that we're not calling update-ca-cert for no reason at the end of the function. I also added an info log about pulling certs from the relation.

bencekov:
To avoid the issues with asyncio that came from the recent breaking changes I pinned pytest-asyncio to version 0.21.1.

metadata.yaml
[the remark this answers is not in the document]
bencekov: Added comments

.wokeignore
bencekov: fixed .wokeignore

tests/integration/test_trusted_certificates.py
```diff
@@ -0,0 +1,99 @@
...
+    # Get pushed cert
+    received_cert_path = f"/usr/local/share/ca-certificates/trusted-ca-cert-{relation_id}-ca.crt"
+    rc, stdout, stderr = await ops_test.juju(
+        "ssh", "--container", "grafana", unit_name, "cat", f"{received_cert_path}"
+    )
+    # Line ends have to be cleaned for comparison
+    received_cert = stdout.replace("\r\n", "\n").rstrip()
+
+    # Get trusted certs
+    trusted_certs_path = "/etc/ssl/certs/ca-certificates.crt"
```
sed-i, line 80: This kind of an integration test is a whitebox test. A blackbox test would be less brittle and would better cover what we want to cover. [Example](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/integration/test_tls_web.py).

I think a better itest would be

```mermaid
graph LR

self-signed-certs ---|tls-certificates| traefik ---|ingress| prometheus
grafana ---|recv-ca-cert| self-signed-certs
grafana ---|grafana-source| prometheus
```

The problem is that for some reason [querying for datasource health](https://grafana.com/docs/grafana/latest/developers/http_api/data_source/#check-data-source-health) doesn't work:

```
$ juju run graf/0 get-admin-password
admin-password: jsbSwIxR6Dg6
url: http://graf-0.graf-endpoints.auth2.svc.cluster.local:3000

$ curl "10.1.166.122:3000/api/datasources" -u admin:jsbSwIxR6Dg6 | jq
[
  {
    "id": 1,
    "uid": "P5DE4FBB56E88771A",
    "orgId": 1,
    "name": "juju_auth2_ee15db9c-9f5d-4594-8b77-3a78862775f4_prom_0",
    "type": "prometheus",
    "typeName": "Prometheus",
    "typeLogoUrl": "public/app/plugins/datasource/prometheus/img/prometheus_logo.svg",
    "access": "proxy",
    "url": "http://prom-0.prom-endpoints.auth2.svc.cluster.local:9090",
    "user": "",
    "database": "",
    "basicAuth": false,
    "isDefault": false,
    "jsonData": {
      "timeInterval": "1m",
      "timeout": 300
    },
    "readOnly": true
  }
]

$ curl "10.1.166.122:3000/api/datasources/uid/P5DE4FBB56E88771A/health" -u admin:jsbSwIxR6Dg6 | jq
{
  "message": "Not found",
  "traceID": ""
}
```

Any ideas?

bencekov: Raised this as an issue.

PR #286, opened 2024-01-05
Use dockerhub image instead of ghcr dev tag

sed-i:
## Issue
Metadata yaml points to ghcr dev tag, which is opaque.

## Solution
Use dockerhub tag.

## Context
- https://github.com/canonical/oci-factory/commits/main/oci/grafana
- https://github.com/canonical/grafana-rock/commits/main/10.2.3


lucabello (approved):
CI's green!

PR #309, opened 2024-03-15
integrations for ceph and migrating to the new charmcraft.yaml

simskij:
_This PR additionally merges our yaml files into `charmcraft.yaml`._

## Issue

Ceph, among other solutions, require embedding of panels and anonymous access.

## Solution

Allow anonymous access and embedding of dashboard panels. Configured through juju config, which defaults to disabled.

## Testing Instructions

- Enable the configs and try to access Grafana.
- Make a dashboard embeddable and try embedding it somewhere.

## Upgrade Notes

N/A


lucabello (approved):

Abuelodelanada (approved):
LGTM.

Shouldn't we have an integration test that checks you can access a dashboard with anonymous access?

simskij:
> LGTM.
> 
> Shouldn't we have an integration test that checks you can access a dashboard with anonymous access?

To me, that feels a little bit like testing the workload, to be honest.

PR #311, opened 2024-03-22
install-plugin action

PietroPasotti:
adds a development action to download and install into grafana a plugin.

Usage:
`juju run grafana install-plugin some-plugin 1.0.0`

The script will grab the plugin from `grafana.com/api/plugins/some-plugin/versions/1.0.0/downloads`, and unzip it into `/var/lib/grafana/plugins`, then restart grafana and the plugin should be there.

Can't test it at the moment because `charmcraft pack` is bork on main.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -3,3 +3,18 @@ get-admin-password:
     Get the dashboard url and initial admin password for the Grafana web interface. Initial
     admin password is generated at charm deployment time. If the password has been changed,
     a notice of that fact will be returned by this action instead.
+
+install-plugin:
+  description: |
+    Installs a grafana plugin.
+  params:
+    name:
+      type: string
+      description: |
+        Name of the plugin. Must be hosted at https://grafana.com/grafana/plugins`.
+        For example "grafana-snowflake-datasource" or "grafana-piechart-panel".
+    version:
+      type: string
+      description: |  
+        Full version of the plugin you want to install. 
+        Must include major, minor, and revision. For example "1.2.3" or "0.0.1"
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,11 +27,13 @@ import secrets
 import socket
 import string
 import time
+import zipfile
 from io import StringIO
 from pathlib import Path
 from typing import Any, Callable, Dict, cast, Optional
 from urllib.parse import urlparse
 import subprocess
+from urllib.request import urlretrieve
 
 import yaml
 from charms.catalogue_k8s.v1.catalogue import CatalogueConsumer, CatalogueItem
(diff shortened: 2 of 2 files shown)


actions.yaml
```diff
@@ -3,3 +3,18 @@ get-admin-password:
...
     a notice of that fact will be returned by this action instead.
+
+install-plugin:
+  description: |
+    Installs a grafana plugin.
+  params:
+    name:
+      type: string
+      description: |
+        Name of the plugin. Must be hosted at https://grafana.com/grafana/plugins`.
```
sed-i, line 14: Would it make more sense to have `url` as the param?
That would lift the `grafana.com` restriction and the `/plugins/{name}/versions/{version}/download` assumption.

Or, if we want to keep the `grafana.com` restriction, then we could use [grafana cli](https://grafana.com/docs/grafana/latest/cli/#plugins-commands) for that, which would be great help with uninstalling (which is currently missing in the PR).

src/charm.py
```diff
@@ -1152,6 +1154,41 @@ def _generate_datasource_config(self) -> str:
         datasources_string = yaml.dump(datasources_dict)
         return datasources_string
 
+    def _on_install_plugin(self, event: ActionEvent) -> None:
+        """Install an official grafana plugin."""
+        name = event.params['name']
+        version = event.params['version']
+        url = f"https://grafana.com/api/plugins/{name}/versions/{version}/download"
+        path = Path('/var/lib/grafana/plugins')
```
Abuelodelanada, line 1162: I would extract path and url to constants

src/charm.py
```diff
@@ -1152,6 +1154,41 @@ def _generate_datasource_config(self) -> str:
...
+        with zipfile.ZipFile(download_path, 'r') as zip_ref:
+            try:
+                zip_ref.extractall(path)
+            except Exception as e:
+                logger.exception(f"cannot unzip {download_path}")
+                event.fail(f"failed to unzip downloaded file at {download_path}")
+                return
+
+        installed_plugins = [o.strip() for o in subprocess.getoutput("grafana cli plugin ls").splitlines()[1:]]
+        download_path.unlink()
```
Abuelodelanada, line 1184: I would put this inside a `try...except` blick

simskij:
This is currently failing checks. Do we want to merge this? If so, we should fix the linting issues. I'm happy having this merged as-is, as it helps with provisioning plugins during initial deployment. Deleting them can be done through the UI, so it doesn't feel as urgent.

lucabello:
@PietroPasotti @simskij @sed-i what do we do with this PR?

lucabello:
Closing this PR because it's stale.

PR #314, opened 2024-03-26
configure database path

### LINKED ISSUE #313: Recreating grafana pods leads to deleting non default organisation and users
### Bug Description

Deleting grafana pod when grafana 9.5.3 is used, leading to removal all custom user and organisations.
We are suspecting that the issue is related to grafana image it is happening with resource revision 67 and 68 but not happening with 63 the charm is latest/stable rev 105

### To Reproduce

1. Deploy cos bundle with the latest stable, make sure that PVs for grafana are created correctly and working
2. Create user and/or organisation through grafana web UI
3. delete grafana pod 'kubectl delete pod -n cos grafana-0`
4. After pod is recreated all users is gone

### Environment

Grafana charm latest/stabel rev 105

### Relevant log output

```shell
I wasn't able to capture logs related to the issue
```

### Additional context

_No response_

Abuelodelanada:
## Issue

This PR fixes #313

## Solution

We add a new config that stores the sqlite3 database in the PVC.

## Testing Instructions

- Run itests
or
- Deploy grafana
- Create a new Organisation
- Delete the pod
- Check that the Organisation is still there.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -112,6 +112,7 @@ CA_CERT_PATH = Path("/usr/local/share/ca-certificates/cos-ca.crt")
 DATABASE = "database"
 PEER = "grafana"
 PORT = 3000
+DATABASE_PATH = "/var/lib/grafana/grafana.db"
 
 # Template for storing trusted certificate in a file.
 TRUSTED_CA_TEMPLATE = string.Template(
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -754,6 +755,16 @@ def _generate_grafana_config(self) -> str:
...
+        else:
+            with StringIO() as data:
+                config_ini = configparser.ConfigParser()
+                config_ini["database"] = {
+                    "type": "sqlite3",
+                    "path": DATABASE_PATH,
+                }
+                config_ini.write(data)
+                data.seek(0)
+                configs.append(data.read())
```
sed-i, line 767: - Shouldn't we render the `path` etc. unconditionally? I.e. why is this section under an `else`?
- Should we have a free function e.g. `dict_to_toml` or `dict_to_ini`?

Abuelodelanada: > * Shouldn't we render the `path` etc. unconditionally? I.e. why is this section under an `else`?

Nop, only if use sqlite as a DB.
This section is under the `else` because the `if` handles when we use MySQL or any DB charm that supports the `db` interface

> 
> * Should we have a free function e.g. `dict_to_toml` or `dict_to_ini`?

We can added it later if we have more of these...

sed-i (approved):

PR #318, opened 2024-04-05
Tracing v2

PietroPasotti:
Bumps tracing interface to v2.


mmkay (approved):

src/charm.py
```diff
@@ -1486,7 +1486,7 @@ def _on_oauth_info_changed(self, event: OAuthInfoChangedEvent) -> None:
     @property
     def tracing_endpoint(self) -> Optional[str]:
         """Tempo endpoint for charm tracing."""
-        return self.tracing.otlp_http_endpoint()
+        return self.tracing.get_endpoint("otlp_http")
```
mmkay, line 1489: I remember what these errors in test logs are!

We need a check on `if self.tracing.is_ready():` - just stumbled upon this while testing loki nginx.

PR #350, opened 2024-10-07
Add terraform module

MichaelThamm:
## Solution

TODO:
- [x] Update the [.gitignore file for terraform files](https://github.com/canonical/terraform-modules/blob/main/.gitignore).
- [x] Call `terraform fmt -check -recursive` as part of CI
  - [x] github workflow (local demo)
  - [x] Tandem PR in observability repo
- [x] Add integrations into module itself (is it possible / sensible?)
- [x] Add a "[constraints](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#constraints)" input var. [Example](https://discourse.charmhub.io/t/pod-priority-and-affinity-in-juju-charms/4091/13).

Refs:
- https://discourse.charmhub.io/t/introducing-the-juju-terraform-provider/6939
- https://github.com/juju/terraform-provider-juju/blob/main/.github/workflows/test_integration.yml
- https://registry.terraform.io/providers/juju/juju/latest/docs#example-usage
- https://github.com/canonical/terraform-modules/tree/main/modules/machine/postgresql-ha

## Testing Instructions
`terraform fmt -check -recursive -diff`


MichaelThamm:
Note that the unit tests are failing due to a breaking change in `ops > 2.16`.
[Here is a PR](https://github.com/canonical/grafana-k8s-operator/pull/352) that addresses it.

terraform/main.tf
```diff
@@ -0,0 +1,13 @@
+resource "juju_application" "grafana" {
+  name = var.app_name
+  # Coordinator and worker must be in the same model
```
sed-i, line 3:
```suggestion
```

terraform/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "app_name" {
+  value = juju_application.grafana.name
+}
```
Gmerold, line 3: What about the `provides` and `requires` outputs? Are you going to add them?

terraform/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_version = ">= 1.5"
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = "~> 0.14.0"
```
Gmerold, line 6: In the Telco Team, we've been working with Terraform for quite some time now and from that perspective, I'd suggest not using a specific version. This quickly leads to compatibility issues in solution modules. You'd better off specifying a range here, i.e. `>= 0.14.0`

Gmerold:
@MichaelThamm Thanks for the update in my old PRs :)

terraform/outputs.tf
MichaelThamm: Just added them

terraform/versions.tf
Gmerold: Looking at the [Juju Terraform provider releases](https://github.com/juju/terraform-provider-juju/tags) I don't think it makes a whole lotta difference ;)

MichaelThamm: The point is that we are willing to receive Juju TF provider updates for minors and patches but for majors we believe it makes sense to hold off on that.

sed-i (approved):

terraform/versions.tf
Gmerold: Do you have a plan for orchestrating TF provider version updates in a way that would not impact the COS Lite solution module (I assume you'll be creating it as well, right?)? Or do you accept the fact that the solution module may be broken if you don't update the provider version in all the application modules at the same time?

PR #354, opened 2024-11-02
Add config option for disabling reporting

### LINKED ISSUE #353: add option --disable-reporting to the command line via charm configuration switch
### Enhancement Proposal

Currently grafana (as well as [grafana-agent](https://github.com/canonical/grafana-agent-operator/issues/197)) calls home (sends anonymous usage statistics) to stats.grafana.org
Our customer does not want any application to call home.

More details available here:
https://grafana.com/docs/alloy/latest/data-collection/

So essentially the charm should have option to --disable-reporting to the command line.

Marking it as field-critical as it required for the closing deployment in 2 weeks.

sed-i:
## Issue
By default, grafana [sends out anonymous usage stats](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#analytics). This may be undesired in hardened environments.

## Solution
Add charm config option to disable reporting.

- In grafana's config file, this option is called `reporting_enabled` ([ref](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#analytics)), while in grafana-agent there's a cli flag called `-disable-reporting`. We want to be consistent across charms, as well as use a positive rather than negative term, for simplicity. So we'll go with `reporting_enabled` everywhere.
- There is asymmetry between `reporting_enabled=true` and `reporting_enabled=false`: when `true`, we do not render any config, so grafana will use default values; when false, we disable both reporting and plugin update checks. This is desirable because:
  - Having a single config option (at least for now) is a simple solution that covers most use cases.
  - If an admin needs to disable reporting, it is also likely they would like to prevent auto checking for plugin updates.
  - In the future we could add another config option for plugin checks without any breaking changes.

In tandem with:
- https://github.com/canonical/grafana-agent-k8s-operator/pull/326
- (Also for Loki: https://grafana.com/docs/loki/latest/configure/#analytics)

Fixes #353.

## Context
- [ADR](https://github.com/canonical/observability/blob/085a923571a3117437be1abb49136c26ac88f567/decision-records/2024-06-27--upstream-telemetry.md)


charmcraft.yaml
```diff
@@ -212,3 +212,14 @@ config:
         automatically deduced from it).
         See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
       type: string
+    phone_home:
```
lucabello, line 215: We're calling this `disable_reporting` in Grafana Agent, it should probably have the same name here

lucabello (approved):

PR #358, opened 2024-11-20
Split endpoints into charm-tracing and workload-tracing

mmkay:

### DIFF
--- a/tests/integration/test_workload_tracing.py
+++ b/tests/integration/test_workload_tracing.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from pathlib import Path
+
+import pytest
+import yaml
+from helpers import oci_image, check_grafana_is_ready
+from test_helpers import deploy_cluster, get_traces_patiently, get_application_ip
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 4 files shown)


tests/integration/helpers.py
```diff
@@ -377,6 +383,117 @@ async def curl(ops_test: OpsTest, *, cert_dir: str, cert_path: str, ip_addr: str
     return stdout
 
 
+async def get_unit_address(ops_test: OpsTest, app_name, unit_no) -> str:
```
michaeldmitry, line 386: We have a duplicate function `unit_address`

tests/integration/test_workload_tracing.py
```diff
@@ -0,0 +1,58 @@
...
+    check_grafana_is_ready,
+    deploy_tempo_cluster,
+    get_traces_patiently,
+    get_application_ip,
+)
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./charmcraft.yaml").read_text())
+app_name = "grafana"
```
michaeldmitry, line 21: should we stick with the other naming conventions of uppercase?

tests/integration/helpers.py
mmkay: Correct, removed a duplicate.

tests/integration/test_workload_tracing.py
mmkay: We definitely should. Fixed.

michaeldmitry (approved):

PR #359, opened 2024-11-25
grafana source uid added to grafana-source interface

PietroPasotti:
## Issue
The grafana source providers need to receive back from grafana the UIDs assigned to them. This, to support cross-referencing functionalities such as https://github.com/canonical/tempo-coordinator-k8s-operator/pull/77.

[Tandem PR in prometheus to forward this data to remote-write integrations](https://github.com/canonical/prometheus-k8s-operator/pull/647)

## Solution
added a `grafana_source_uids: Dict[unit_name: source_uid]` field to the application databag on the grafana side.

All changes are additive and therefore backwards-compatible.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -432,7 +432,8 @@ class GrafanaSourceProvider(Object):
     def get_source_uids(self) -> Dict[str, Dict[str, str]]:
         """Get the datasource UID(s) assigned by the remote end(s) to this datasource.
 
-        Returns a mapping from remote application names to unit names to datasource uids."""
+        Returns a mapping from remote application names to unit names to datasource uids.
+        """
         uids = {}
         for rel in self._charm.model.relations.get(self._relation_name, []):
             if not rel:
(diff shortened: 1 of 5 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -551,6 +563,13 @@ def _on_grafana_peer_changed(self, _: RelationChangedEvent) -> None:
         self.on.sources_changed.emit()  # pyright: ignore
         self.on.sources_to_delete_changed.emit()  # pyright: ignore
 
+    def _publish_source_uids(self, rel: Relation, uids: Dict[str, str]):
+        """Share the datasource UIDs back to the datasources.
+
+        Assumes only leader unit will call this method
+        """
+        rel.data[self._charm.app]["datasource_uids"] = json.dumps(uids)
```
mmkay, line 571: Should we have one more leader check just in case someone misuses the method?

mmkay (approved):
Just a minor comment, otherwise looks good.

lib/charms/grafana_k8s/v0/grafana_source.py
michaeldmitry: hmm I think in this case, I'd want it error out loudly rather than silently not doing anything

michaeldmitry: plus its intended for private use

PR #361, opened 2024-12-02
Datasource interface tests

PietroPasotti:
Interface test fixture for the grafana_datasource interface tests.

See: https://github.com/canonical/charm-relation-interfaces/pull/204


tests/scenario/conftest.py
```diff
@@ -6,29 +7,22 @@
...
+        stack.enter_context(patch("socket.gethostbyname", new=lambda *args: "1.2.3.4"))
+        stack.enter_context(
+            patch.multiple(
+                "charm.KubernetesComputeResourcesPatch",
+                _namespace="test-namespace",
+                _patch=lambda *_a, **_k: True,
+                is_ready=lambda *_a, **_k: True,
+            )
+        )
+        stack.enter_context(patch.object(GrafanaCharm, "grafana_version", "0.1.0"))
```
sed-i, line 27: Nit: reduce duplication by using a for loop like in the [docs](https://docs.python.org/3.8/library/contextlib.html#contextlib.ExitStack).

tests/interface/conftest.py
sed-i, line 1: Practically, scenario tests and interface tests are all unit tests.
This PR duplicates the code for the patches we apply.
Can we move all this under `unit` instead?

PietroPasotti: not without one more PR to charm-relation-interfaces, because by default, it's looking for interface tests under tests/interface.

tests/scenario/conftest.py
PietroPasotti: ugh, a list comprehension as a for loop... I don't mind using a for loop though.

lucabello:
Closing this as it's 1.5 years old.

PR #362, opened 2024-12-04
allow empty datasource_uids field without raising

PietroPasotti:
return an empty object instead of raising if the remote end uses an old interface


michaeldmitry (approved):
linting is failing

PR #364, opened 2024-12-09
added grafana uid

PietroPasotti:
Adds grafana's own fingerprint UID to the grafana_source relation databag like specified in https://github.com/canonical/charm-relation-interfaces/pull/211

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -164,7 +164,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 23
+LIBPATCH = 24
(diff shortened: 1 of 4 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -154,6 +154,8 @@ def __init__(self, *args):
 )
 from ops.model import Relation
 
+from charms.observability_libs.v0.juju_topology import JujuTopology
```
michaeldmitry, line 157: BTW `charms.observability_libs.v0.juju_topology` is deprecated. Should we be using the one provided by `cosl?`

PietroPasotti: mmm, I'm not sure, it's out of scope for this PR and I'm afraid we're not the only ones using this charm lib, do we want to force everyone to switch dependencies?

PR #369, opened 2024-12-18
tests: relation_broken.emit() takes a relation, not a relation ID

tonyandrewmeyer:
## Issue

A broken test (not noticeable at the moment, although it's not doing the right thing, but noticeable with the snapshot changes in https://github.com/canonical/operator/pull/1372).

## Solution

Pass the correct argument to `relation_broken.emit()`

## Context

ops?

## Testing Instructions

`tox -e unit`

## Upgrade Notes

N/A


mmkay (approved):
Thank you for the contribution @tonyandrewmeyer!

PR #370, opened 2025-01-09
Refactor dashboard utils

sed-i:
This PR is a small, scoped change, extracted from #363.
This way #363 will be easier to review, and the changes in this PR are trivial.

- Use `cosl.LZMABase64` instead of `_encode_dashboard_content`, `_decode_dashboard_content`.
- Namespace the following free functions under `class CharmedDashboard`. Github's diff algorithm produces a messy diff. In the attached diff ([report.pdf](https://github.com/user-attachments/files/18366996/report.pdf)) it's easier to see that the difference is just indentation.
  - `_convert_dashboard_fields`
  - `_replace_template_fields`
  - `_template_panels`
  - `_inject_labels`
  - `_modify_panel`

PR #363, opened 2024-12-06
Automatically overwrite dashboard UIDs with a prescribed naming pattern, on the provider side

sed-i:
## Issue
- https://github.com/canonical/observability/blob/main/decision-records/2024-08-29--dashboard-uid-collision.md

## Solution
Use shake_256 to generate a prescribed uid, overwriting any pre-existing uid.

In tandem with:
- https://github.com/canonical/cos-lib/pull/114
- https://github.com/canonical/grafana-agent-operator/pull/224
- https://github.com/canonical/grafana-agent-k8s-operator/pull/337
- https://github.com/canonical/cos-proxy-operator/pull/174

## Context
- Afaict, the only user of the `.add_dashboard` method is https://github.com/canonical/cos-registration-server-k8s-operator/blob/6e156f0b1367f7cc6ce12d89f6a367bf491fea7f/src/charm.py#L240.
- [Dahsboard templates](https://github.com/search?q=org%3Acanonical+path%3A.tmpl+grafana&type=code) ([example](https://github.com/canonical/cos-proxy-operator/blob/2835a6b79b30cc5272237f31ae2d1034859ee1a6/src/grafana_dashboards/impi-sensor.json.tmpl#L6)) are rendered on the consumer side (grafana itself). The templates themselves are still valid json, so it's ok to `json.loads` them in the provider side.

## Testing Instructions
First, prepare the charms locally: 
1. Copy the grafana_dashboard lib from this PR into grafana-agent-operator from https://github.com/canonical/grafana-agent-operator/pull/224.
2. Copy cos-agent lib from https://github.com/canonical/grafana-agent-operator/pull/224 into zookeeper.
3. In zookeeper, update deps using the following commands. Then remove all the `--hash` from requirements.txt:

```
poetry update
poetry lock
poetry install
poetry export -f requirements.txt --output requirements.tx
```

4. Pack and deploy all charms as follows:

```mermaid
graph LR
subgraph lxd
zookeeper --- grafana-agent
zookeeper2 --- grafana-agent
end
subgraph microk8s
grafana-agent --- grafana
end
```

```yaml
bundle: kubernetes
applications:
  graf:
    charm: ./grafana-k8s_ubuntu-20.04-amd64.charm
    scale: 1
    trust: true
--- # overlay.yaml
applications:
  graf:
    offers:
      graf:
        endpoints:
        - grafana-dashboard
        acl:
          admin: admin
```
```yaml
default-base: ubuntu@22.04/stable
saas:
  graf:
[...]

PR #373, opened 2025-01-15
Refactor dashboard utils (part 2)

sed-i:
Following up on #370, as part of simplifying #363.
This PR introduces no functional changes.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -890,6 +890,51 @@ class CharmedDashboard:
         panel["targets"] = targets
         return panel
 
+    @classmethod
+    def _content_to_dashboard_object(
+        cls,
+        *,
+        charm_name,
+        content: str,
+        juju_topology: dict,
+        inject_dropdowns: bool = True,
+        dashboard_alt_uid: Optional[str] = None,
+    ) -> Dict:
+        """Helper method for keeping a consistent stored state schema for the dashboard and some metadata.
+
+        Args:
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -890,6 +890,51 @@ def _modify_panel(cls, panel: dict, topology: dict, transformer: "CosTool") -> d
...
+    def _generate_alt_uid(cls, charm_name: str, key: str) -> str:
+        """Generate alternative uid for dashboards.
+
+        Args:
+            charm_name: The name of the charm (not app; from metadata).
+            key: A string used (along with charm.meta.name) to build the hash uid.
+
+        Returns: A hash string.
+        """
+        raw_dashboard_alt_uid = "{}-{}".format(charm_name, key)
```
Abuelodelanada, line 935: We can safely use f-strings since it was introduced in python 3.6 (Ubuntu 18.04), and [there is no charm based in a previous version using this lib.
](https://charmhub.io/integrations/grafana_dashboard)
```suggestion
        raw_dashboard_alt_uid = f"{charm_name}-{key}"
```

Abuelodelanada (approved):
LGTM!

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: Good point. Let's keep this in mind for a [pyupgrade](https://github.com/asottile/pyupgrade) PR.

PR #376, opened 2025-01-16
fix: set the correct permissions for cos-tool in tox.ini

lucabello:
This issue was not appearing before because we were running `chmod` in the charm libraries. Now that we removed that, we should fix it in `tox.ini`.


sed-i (approved):

dimaqq:
Whenever I come to complain about something in your repos, I find that it has recently been fixed!

Keep up the good job :)

PR #387, opened 2025-02-05
refactor: move the grafana UID definition to the charm from grafana_source

ca-scribner:
The grafana_source relation sends a unique string identifying this grafana.  Previously, this string was created in the grafana_source library, making it a bit awkward to use in other libraries.

This PR proposes a change where this identifier is created/owned by the grafana charm rather than the datasource lib.  Outwardly, this should cause no change in function.  The motivation for this change is to use this same identifier in the forthcoming grafana-metadata relation

The grafana_uid is meant for identifying an application of charmed grafana, so it should be universally unique (unique even across different models, deployments, etc).  Given that, it really is owned by grafana the application, and the related lib merely *uses* that ID.  

This change introduces a breaking change in the library API due to an additional required parameter.  The libapi has not been bumped here because grafana-k8s is the only operator that consumes this (and because, if there is someone else consuming it, the change in api will result in an obvious bug rather than a hidden bug, so the risk of real damage is low).  

## Testing Instructions
Should be covered by existing CI, apart from a slight fix in unit tests

## Upgrade Notes
Should not have any functional effect on the charm

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -512,7 +512,7 @@ class GrafanaSourceConsumer(Object):
         Args:
             charm: a :class:`CharmBase` instance that manages this
                 instance of the Grafana source service.
-            grafana_uid: a unique identifier for this Grafana datasource.
+            grafana_uid: an identifier uniquely identifying this grafana-k8s application.
             relation_name: string name of the relation that is provides the
                 Grafana source service. It is strongly advised not to change
                 the default, so that people deploying your charm will have a
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -512,7 +512,7 @@ class GrafanaSourceConsumer(Object):
         Args:
             charm: a :class:`CharmBase` instance that manages this
                 instance of the Grafana source service.
-            grafana_uid: an identifier uniquely identifying this grafana-k8s application.
+            grafana_uid: an unique identifier for this grafana-k8s application.
             relation_name: string name of the relation that is provides the
                 Grafana source service. It is strongly advised not to change
                 the default, so that people deploying your charm will have a
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -162,7 +162,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 24
+LIBPATCH = 25
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -504,6 +504,7 @@ class GrafanaSourceConsumer(Object):
     def __init__(
         self,
         charm: CharmBase,
+        grafana_uid: str,
         relation_name: str = DEFAULT_RELATION_NAME,
     ) -> None:
         """A Grafana based Monitoring service consumer, i.e., the charm that uses a datasource.
@@ -511,6 +512,7 @@ class GrafanaSourceConsumer(Object):
         Args:
             charm: a :class:`CharmBase` instance that manages this
                 instance of the Grafana source service.
+            grafana_uid: a unique identifier for this Grafana datasource.
             relation_name: string name of the relation that is provides the
                 Grafana source service. It is strongly advised not to change
                 the default, so that people deploying your charm will have a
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -1623,6 +1623,16 @@ def _push_sqlite_static(self):
...
 
+    @property
+    def unique_name(self):
+        """Returns a unique identifier for this application."""
+        return "juju_{}_{}_{}_{}".format(
+            self.model.name,
+            self.model.uuid,
+            self.model.app.name,
+            self.model.unit.name.split("/")[1],  # type: ignore
+        )
```
sed-i, line 1634: Can you elaborate on the higher level goals? I'm thinking:
- It is the grafana-source lib that gets to decide how a [source uid](https://github.com/canonical/grafana-k8s-operator/blob/30b39fc23710b90cc866f48132e1b2e75b710b34/lib/charms/grafana_k8s/v0/grafana_source.py#L578C18-L578C29) looks like, for the purpose of this relation.
- Why do you need a `juju_` prefix here?
- Is there a juju topology method that produces what you want?
- It's a neat idea for all charms in general to have a uuid. I wonder what CharmTech think about adding a `self.unit.uuid` property with exactly this.

ca-scribner: @PietroPasotti can confirm, but afaik the goal of this ID was to allow a related charm to say "I'm related to *the grafana with this unique id*", which is important when communicating where you can find tempo's dashboards.  To that end, he crafted an identifier that should be unique to fit that purpose.  juju actually gives applications a UID of its own, but I think they're inaccessible to charms (asked about that [in matrix](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$bbW9G7bJOM968p_ZJQ1-IOoMyMf2k97BZkZIB6o-8xc?via=ubuntu.com&via=matrix.org&via=xentonix.net))

As for who "owns" the IDs, the datasource_exchange sends both a grafana ID (for the grafana instance, the ID that this PR affects) and datasource IDs (for each datasource - unchanged in this PR).  In my view, the grafana ID seems like it should be owned by the charm itself, not the datasource lib.  That lets it then easily be observed by others, for example through the forthcoming `grafana-metadata` relation (#388)

sed-i: I'm concerned that this "unit uuid" we're crafting might be short sighted in the long term. Let's check with the juju team?

ca-scribner: yeah it does feel like something that could be provided by Juju and that would be better, though in the short term I'm +1 for making our own until they have something suitable

ca-scribner: Confirmed with juju - they provide no alternatives for this.  They also questioned why the unit number would be in this ID but otherwise suggested we do essentially this

Checked with Pietro and he mentioned he copied or nearly copied this from the juju topology lib

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -504,13 +504,15 @@ class GrafanaSourceConsumer(Object):
     def __init__(
         self,
         charm: CharmBase,
+        grafana_uid: str,
         relation_name: str = DEFAULT_RELATION_NAME,
```
sed-i, line 508: Now that we know juju 4 will come with app + unit uid, we should probably prepare for it by designing the API more carefully:
```suggestion
        relation_name: str = DEFAULT_RELATION_NAME,
        *,
        grafana_uid: Optional[str] = None,
```

and when it is None then use the previous logic.

ca-scribner: I talked with Ian about this in person and, presently, the juju 4 app UID wasn't planned to be exposed to charms.  That could change though.

ca-scribner: To make sure I understand your proposal here: the lib should optionally accept `grafana_uid`, and if it is not provided then the lib should create the ID itself like it has done prior to this change?

I can do it that way, but I recommend against letting the UID be optional.  That will make it easier to accidentally misconfigure things.  Is there a case where having this as optional is important?  I can't think of one.

ca-scribner: Clarifying the comment above a bit more... To me, ownership of this ID is in the charm rather than in the lib, so that's why I'd see this as a required argument passed from charm to lib

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -504,13 +504,15 @@ class GrafanaSourceConsumer(Object):
...
     ) -> None:
         """A Grafana based Monitoring service consumer, i.e., the charm that uses a datasource.
 
         Args:
             charm: a :class:`CharmBase` instance that manages this
                 instance of the Grafana source service.
+            grafana_uid: an identifier uniquely identifying this grafana-k8s application.
```
ca-scribner, line 515:
```suggestion
            grafana_uid: an unique identifier for this grafana-k8s application.
```

sed-i (approved):

lib/charms/grafana_k8s/v0/grafana_source.py
ca-scribner: we discussed offline and decided to keep grafana_uid as a position arg.  See this PRs description for more details

src/charm.py
ca-scribner: Left this as is, although probably it makes sense in future to remove the unit number from this ID

PR #388, opened 2025-02-05
feat: add grafana-metadata relation

ca-scribner:
## Issue
grafana-k8s does not have a simple relation that someone can consume when they need access to the grafana dashboard.

## Solution
This PR implements the `grafana-metadata` relation with schema: 

```python
class GrafanaMetadataAppData(BaseModel):
    """Data model for the grafana-info interface."""

    ingress_url: str = Field(description="The ingress URL.")
    internal_url: str = Field(description="The URL for connecting to the prometheus api from inside the cluster.")
    grafana_uid: str = Field(description="The UID of this Grafana application.")
```

This will enable connecting Grafana to Kiali or other consumers of the dashboard.

Because the implementation of this library is identical to others being created for prometheus, istio, and tempo, we've implemented most of the logic in a package `charm-relation-building-blocks` ([initial pr here](https://github.com/canonical/charm-relation-building-blocks/pulls)).  The library code implemented in the grafana-k8s-operator repo is just a concrete implementation of that relation.  

## Context
This PR is part of efforts to enable consuming grafana in Kiali.

Closes https://github.com/canonical/kiali-k8s-operator/issues/27

## Todo:
* [ ] This PR is based on #387.  When that PR merges, we will rebase so their commit leaves the diff.

## Testing Instructions
Unit/Integration tests should cover the changes.  If you want to manually test, use the included grafana-metadata-requirer-tester to relate to grafana.

## Upgrade Notes
This is additive - shouldn't affect upgrades.


lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+        self._charm = charm
+        self._data = GrafanaMetadataAppData(ingress_url=ingress_url, internal_url=internal_url, grafana_uid=grafana_uid)
+        self._relation_name = relation_name
+
+    def _get_relations(self):
+        """Return the applications related to us under the monitored relation."""
+        return self._charm.model.relations.get(self._relation_name, ())
+
+    def send_data(self):
```
dstathis, line 245: I would maybe call this method something like update relation data to more accurately reflect what happens. Not a hard request by any means. Just an idea.

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
+"""grafana_metadata.
+
+This implements provider and requirer sides of the grafana-metadata interface, which is used to communicate information
+about an Grafana installation such as its url and UID.
```
PietroPasotti, line 4:
```suggestion
about a Grafana installation such as its internal and external urls and UID.
```

I'd like to see clearly stated somewhere that the semantics of the interface is: "who am I, and how do you talk to me"

PietroPasotti: also, is this meant to be scalable to applications that have multiple queriable urls?
If one day we solve scaling up grafana, could there be urls for multiple units in there?

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+requires:
+  grafana-metadata:
+    # The example below uses the API for when limit=1.  If you need to support multiple related applications, remove
+    # this and use the list-based data accessor method.
+    limit: 1
+    interface: grafana_metadata
+```
+
+To handle the relation events in your charm, use `GrafanaMetadataRequirer`.  That object handles all relation events for
+this relation, and emits a `DataChangedEvent` when data changes the charm might want to react to occur.  To set it up,
```
PietroPasotti, line 22:
```suggestion
this relation, and emits a `DataChangedEvent` that the charm can observe.  To set it up,
```

as a non-native speaker, this sentence jumbled me up quite seriously

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+instantiate an `GrafanaMetadataRequirer` object in your charm's `__init__` method and observe the `DataChangedEvent`:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        # Create the GrafanaMetadataRequirer instance, providing the relation name you've used
+        self.grafana_metadata = GrafanaMetadataRequirer(self, "grafana-metadata")
+        self.framework.observe(self.grafana_metadata.on.data_changed, self.do_something_with_metadata)
```
PietroPasotti, line 31:
```suggestion
        self.framework.observe(self.grafana_metadata.on.data_changed, self._on_grafana_metadata_data_changed)
```

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+objects:
+
+```python
+class FooCharm(CharmBase):
+    ...
+    # If using limit=1
+    def do_something_with_metadata(self):
+        # Get exactly one related application's data, raising if more than one is available
+        # note: if not using limit=1, see .get_data_from_all_relations()
+        metadata = self.grafana_metadata.get_data()
```
PietroPasotti, line 44: thought, how about this API instead:
```
if not self.grafana_metadata.is_ready():
    self.log(...)
    return
internal_url = self.grafana_metadata.internal_url
```

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+To manage sending data to related applications in your charm, use `GrafanaMetadataProvider`.  Note that
+`GrafanaMetadataProvider` *does not* manage any events, but instead provides a `send_data` method for sending data to
+all related applications.  Triggering `send_data` appropriately is left to the charm author, although generally you want
+to do this at least during relation_joined and leader_elected events.  An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.grafana_metadata = GrafanaMetadataProvider(
```
PietroPasotti, line 71: fine with passing the urls and uid in the constructor, but I'd also add a way to do this later, e.g. in a reconcile method.
something like `self.grafana_metadata.set_data(uid, external_url...)

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+PYDEPS = ["pydantic>=2"]
+
+log = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "grafana-info"
+
+
+class GrafanaMetadataAppData(BaseModel):
+    """Data model for the grafana-info interface."""
```
PietroPasotti, line 109:
```suggestion
    """Data model for the grafana-metadata interface."""
```
do a thorough search for `grafana-info`, i've seen it elsewhere too.

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+    ingress_url: str = Field(description="The ingress URL.")
```
PietroPasotti, line 111:
```suggestion
    ingress_url: pydantic.AnyUrl = Field(description="The ingress URL.")
``` 
perhaps?

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+    internal_url: str = Field(description="The URL for connecting to the prometheus api from inside the cluster.")
+    grafana_uid: str = Field(description="The UID of this Grafana application.")
+
+
+class DataChangedEvent(EventBase):
```
PietroPasotti, line 116: if you want to do an event-based API, I think it's a good idea to assign to the event all the metadata so one can do:

```
def _on_event(self, e:DataChangedEvent):
    uid=e.grafana_uid
    url=e.grafana_url
```

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+class GrafanaMetadataRequirer(Object):
+    """Class for handling the receiver side of the grafana-info relation."""
+
+    on = ReceiverCharmEvents()  # type: ignore[reportAssignmentType]
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str,
+        refresh_event: Optional[Union[BoundEvent, List[BoundEvent]]] = None,
```
PietroPasotti, line 135: I regret the moment this ever became a pattern. IMO this is a better signature
```suggestion
        refresh_events: Sequence[BoundEvent] = (),
```

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class ReceiverCharmEvents(CharmEvents):
+    """Events raised by the data receiver side of the interface."""
+
+    data_changed = EventSource(DataChangedEvent)
+
+
+class GrafanaMetadataRequirer(Object):
+    """Class for handling the receiver side of the grafana-info relation."""
```
PietroPasotti, line 127:
```suggestion
    """The requirer side of the grafana-metadata relation."""
```

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        if isinstance(refresh_event, BoundEvent):
+            refresh_event = [refresh_event]
+        for ev in refresh_event:
+            self.framework.observe(ev, self.on_relation_changed)
+
+        self.framework.observe(
+            self._charm.on[self._relation_name].relation_changed, self.on_relation_changed
+        )
+
+    def __len__(self):
```
PietroPasotti, line 161: ? Never seen this in other relation wrappers. What's the use case?
Is it not more explicit to have `len(grafana_metadata.relations)` if you need that in the charm?

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        """Return the number of related applications."""
+        return len(self.get_relations())
+
+    def on_relation_changed(self, _: EventBase) -> None:
+        """Handle when the remote application data changed."""
+        self.on.data_changed.emit()
+
+    def get_relations(self):
```
PietroPasotti, line 169: usually this is a `relations` property

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        """Return the relation instances for applications related to us on the monitored relation."""
+        return self._charm.model.relations.get(self._relation_name, ())
+
+    def get_data(self) -> Optional[BaseModel]:
```
PietroPasotti, line 173: instead of having two methods for this, consider the signature:
`get_data(self, *relation:Optional[Relation]=None) -> Sequence[GrafanaMetadata]:`

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
+        raw_data_dict = {str(k): v for k, v in raw_data_dict.items()}
+
+        return GrafanaMetadataAppData.model_validate_json(json.dumps(raw_data_dict))  # type: ignore
+
+    def get_data_from_all_relations(self) -> List[BaseModel]:
+        """Return a list of data objects from all relations."""
+        relations = self.get_relations()
+        info_list = []
+        for i, relation in enumerate(relations):
```
PietroPasotti, line 200: what are we enumerating for?

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+            data_dict = relation.data.get(relation.app)
+            if not data_dict:
+                info_list.append(None)
+                continue
+
+            # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
+            data_dict = {str(k): v for k, v in data_dict.items()}
+            info_list.append(GrafanaMetadataAppData(**data_dict))
+        return info_list
```
PietroPasotti, line 209: so, this could contain some `None`s? see line 202.

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class GrafanaMetadataProvider(Object):
+    """Class for handling the Provider side of the grafana-info relation."""
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        grafana_uid: str,
```
PietroPasotti, line 218: I vote we make these optional to allow for late-setting the data in a reconciler method. see other APIs that do the same, such as ingress, grafana datasource...

lib/charms/grafana_k8s/v0/grafana_metadata.py
PietroPasotti: in traefik it's `publish_`

src/charm.py
```diff
@@ -331,6 +333,19 @@ def __init__(self, *args):
...
 
+        # -- grafana-metadata relation handling
+        self.grafana_metadata = GrafanaMetadataProvider(
+            charm=self,
+            grafana_uid=self.unique_name,
+            ingress_url=self.external_url,
+            internal_url=self.internal_url,
+            relation_name=GRAFANA_METADATA_RELATION_NAME,
+        )
+        self.framework.observe(self.on.leader_elected, self._send_grafana_metadata)
```
PietroPasotti, line 344: better practice:
 
 ```suggestion
        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
```

**Review by PietroPasotti — COMMENTED** [review]
not codeowner, so not blocking, but have a bunch of thoughts about this

File: src/charm.py [hunk 19]
```diff
@@ -1641,6 +1656,10 @@ def unique_name(self):
             self.model.unit.name.split("/")[1],  # type: ignore
         )
 
+    def _send_grafana_metadata(self, _):
+        """Send metadata to related applications on the grafana-metadata relation."""
+        self.grafana_metadata.send_data()
```
dstathis, line 1661: I think this sets app data right? So should be leader guarded.

lib/charms/grafana_k8s/v0/grafana_metadata.py
dstathis: If one day we solve scaling up Grafana, we would presumably switch to ingress-per-app right? At that point you would still want only one URL.

But that does bring up the question of what internal address to supply when we scale up grafana now. Do we need to supply the leaders address rather than the service address?

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: as a native speaker, this sentence jumbled me up as well :D

src/charm.py
dstathis: Can you elaborate as to why this is bad practice

PietroPasotti: https://docs.google.com/document/d/1G62PosrObvmQY5KbxvqaxByojlhDxrmNtcbPS39YbaY/edit?tab=t.0#bookmark=id.20k0qpk5irc9 

and that's what we do in all modern charms

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: That works for me

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Yeah at one point these were combined but it shifted more work to the user

If someone sets `limit: 1` in their charmcraft.yaml (probably common for using this relation), their UX is ugly:
```python
try:
  data = grafana_metadata.get_data()[0]
except KeyError:
  data = None
```

instead of
```python
data = grafana_metadata.get_data()
```

I'm not hard set on keeping these split, but it seems like a convenience we should provide to the user

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Yep this got refactoed out.  Will remove

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,250 @@
...
+        if not raw_data_dict:
+            return None
+
+        # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
+        raw_data_dict = {str(k): v for k, v in raw_data_dict.items()}
+
+        return GrafanaMetadataAppData.model_validate_json(json.dumps(raw_data_dict))  # type: ignore
+
+    def get_data_from_all_relations(self) -> List[BaseModel]:
+        """Return a list of data objects from all relations."""
```
ca-scribner, line 197: This should explain the structure of the return, as discussed [here](https://github.com/canonical/grafana-k8s-operator/pull/388/files#r1952468788)

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: hahaha I agree with you, but went for consistency with the others.  If others are onboard too I'm happy to change

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Thinking on it more, its unclear whether `len(grafana_metadata)` will return number or relations or number of metadata's.  All the more reason to remove this

src/charm.py
ca-scribner: Good catch.  Fixed

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: I think this thread can be collapsed into [this one](https://github.com/canonical/grafana-k8s-operator/pull/388#discussion_r1952469766) so I'll resolve this but reply in the other one

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: the `GrafanaMetadataProvider` requires these to function - making them optional feels like a lie no?  

In cases where a reconciler should set these attributes, why instantiate the GrafanaMetadataProvider in `__init__` at all?  Would it work better to just construct the entire object when the data is available in the reconciler?

Taking a step back, why do we instantiate these in `__init__` at all?  I guess its to accommodate libs that should handle events?  Maybe we just shouldn't do that here

lib/charms/grafana_k8s/v0/grafana_metadata.py
PietroPasotti: how about providing two wrappers, one for the 'limit 1 case' and one for the 'no limit' one?
`GrafanaMultiMetadata` :D

lib/charms/grafana_k8s/v0/grafana_metadata.py
PietroPasotti: indeed. I'd say go all-in with reconcile only.

lib/charms/grafana_k8s/v0/grafana_metadata.py
PietroPasotti: well as it stands, the return type annotation is wrong.
how about returning a mapping from Relations to Metadatas?

That way the caller can choose:
use `filter(result, None)` if they only care about the non-None metadatas
use the Relations that are mapped to Nones if the intention is to tell the user "these applications are misbehaving"

src/charm.py
```diff
@@ -1641,6 +1656,12 @@ def unique_name(self):
             self.model.unit.name.split("/")[1],  # type: ignore
         )
 
+    def _send_grafana_metadata(self, _):
+        """Send metadata to related applications on the grafana-metadata relation."""
+        if not self.unit.is_leader():
+            return
+        self.grafana_metadata.send_data()
```
sed-i, line 1663: I'm confused about the reconciler convo above.
It seems like this lib could be reduced to `TheLib.publish(relname, uid, url1, url2)`, and call it from the reconciler method (or the constructor).

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: @PietroPasotti do you mean that we should publish this in charm relation interfaces (and can add a link in the docstring), or something more detailed here?  I'm +1 for that

@dstathis I think we've talked about how we wouldn't scale up grafana (because we'd charm [insert app that I dont remember the name of] instead?)

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,254 @@
...
+
+### Requirer
+
+To add this relation to your charm as a requirer, add the following to your `charmcraft.yaml` or `metadata.yaml`:
+
+```yaml
+requires:
+  grafana-metadata:
+    # The example below uses the API for when limit=1.  If you need to support multiple related applications, remove
+    # this and use the list-based data accessor method.
```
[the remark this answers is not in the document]
ca-scribner: Yes, grafana owns this lib.  

Sorry I don't understand what you mean.  Do you have an example you can link from another lib?

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Reworded and I think it should be clearer, but please reopen if I'm wrong.

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,254 @@
...
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+PYDEPS = ["pydantic>=2"]
+
+log = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "grafana-info"
```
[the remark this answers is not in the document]
ca-scribner: I think I've got everything switched to `grafana-metadata` now

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: good call.  Done, but with `AnyHttpUrl` (scheme must be http or https) as I think that's a better fit.  Please reopen if you disagree!

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class GrafanaMetadataAppData(BaseModel):
+    """Data model for the grafana-info interface."""
+
+    ingress_url: str = Field(description="The ingress URL.")
+    internal_url: str = Field(description="The URL for connecting to the prometheus api from inside the cluster.")
```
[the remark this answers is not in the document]
ca-scribner: tbh I don't think we know.  I'm not convinced we should provide more than one at all, but @dstathis makes good arguments about how Juju doesn't help us here so we need these hacks...

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,254 @@
...
+    grafana_uid: str = Field(description="The UID of this Grafana application.")
+
+
+class DataChangedEvent(EventBase):
+    """Charm Event triggered when the relation data has changed."""
+
+
+class ReceiverCharmEvents(CharmEvents):
+    """Events raised by the data receiver side of the interface."""
```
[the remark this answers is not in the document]
ca-scribner: No, that's leftover from when this was generic.  Changed it now to Requirer

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Done, and the same change for the provider

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: So are we agreed we can go with `refresh_events` (which seems more correct, but is inconsistent with some past libs like [grafana-source](https://github.com/canonical/grafana-k8s-operator/blob/13fa90ed7d5304ce26eaba17f0dff3c2831c07a8/lib/charms/grafana_k8s/v0/grafana_source.py#L326)?

As for the type in @PietroPasotti's suggestion, I think his suggestion looks good.  The type we accept is a Sequence (list, mutable or not, is fine for our purposes) but then we default to an empty tuple (which is a sequence).  @sed-i wdyt?

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: removed

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,254 @@
...
+            self._charm.on[self._relation_name].relation_changed, self.on_relation_changed
+        )
+
+    def __len__(self):
+        """Return the number of related applications."""
+        return len(self.get_relations())
+
+    def on_relation_changed(self, _: EventBase) -> None:
+        """Handle when the remote application data changed."""
+        self.on.data_changed.emit()
```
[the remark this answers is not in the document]
ca-scribner: I think there's a good case for not needing the `charm` arg at all (and that @PietroPasotti would be onboard with that), although I don't see how this pattern relates.  I'll make a cafe topic for this to discuss more as I think it'll make sense in person

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: I like that way better.  So then we'd have:

* `GrafanaMetadataRequirer`: for limit=1.  `get_data()` returns `Optional[GrafanaMetadataAppData]`
* `GrafanaMultiMetadataRequirer`: for any limit.  `get_data()` returns `List[GrafanaMetadataAppData]`, which could be an empty list

I'm less sure about the names :D But I like the UX.

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: I think the return type annotation is correct?  It'll return a list of `GrafanaMetadataAppData`'s, and that could be an empty list

returning a mapping could work.  Maybe that's just the best way.  Requiring someone to use a `filter` for something this common feels a bit underwhelming as a UX though.

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Added a topic to our next cafe that should cover this.  Will come back after that

src/charm.py
ca-scribner: Yes, it could be made functional.  Need to pass the relation objects not just the `relname`, but essentially what you're saying.

charmcraft.yaml
```diff
@@ -91,6 +91,10 @@ provides:
   metrics-endpoint:
     interface: prometheus_scrape
     description: Links to grafana's own `/metrics` endpoint.
+  grafana-metadata:
+    interface: grafana_metadata
+    description: |
+      Provides metadata about the grafana instance, such as its URL and ID.
```
sed-i, line 97: Add more info here targeted at devs? E.g. the limit thing.

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Based on today's team discussion, I think we're moving away from events on this lib, so this'll get refactored away

lib/charms/grafana_k8s/v0/grafana_metadata.py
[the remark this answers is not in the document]
ca-scribner: Sorry I still don't follow :)  This information is saying that for any charm being a requirer to this interface, here is an example of retrieving data if you set limit=1.  I dont know why this would go into grafana's relation description, shouldn't it just be a comment in the docs here?

lib/charms/grafana_k8s/v0/grafana_metadata.py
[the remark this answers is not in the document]
ca-scribner: based on recent video chat, we decided on sending `ingress_url` and `direct_url`

charmcraft.yaml
ca-scribner: Responded in [this thread](https://github.com/canonical/grafana-k8s-operator/pull/388/files#r1954363455)

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: This has been implemented for limit=1 for now, and when we get to limit!=1 we'll implement it with Pietro's suggestion

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: reconciled this code down to reconcile only!

src/charm.py
ca-scribner: left as the fan-out for now.

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: This got refactored out anyway

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Yeah, this got refactored out so I'm resolving the thread

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: Closing as this got refactored out

lib/charms/grafana_k8s/v0/grafana_metadata.py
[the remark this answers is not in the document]
ca-scribner: libs now do not take `charm` as an arg

lib/charms/grafana_k8s/v0/grafana_metadata.py
ca-scribner: removed

src/charm.py
ca-scribner: after a cafe discussing it, we decided to go with simple classes that accept `RelationMapping` not `CharmBase`, and if we want to change later we can refactor

src/charm.py
```diff
@@ -1641,6 +1648,28 @@ def unique_name(self):
...
+
+        grafana_metadata = GrafanaMetadataProvider(
+            relations=self.model.relations,
+            grafana_uid=self.unique_name,
+            ingress_url=external_url,  # pyright: ignore
+            direct_url=internal_url,  # pyright: ignore
+            app=self.app,
+            relation_name="grafana-metadata",
+        )
+        grafana_metadata.publish()
```
sed-i, line 1671: Having this here looks awesome!

tests/integration/test_grafana_metadata.py
sed-i, line 1: What is the delta in coverage the itest adds to the utest?

sed-i (approved):

lib/charms/grafana_k8s/v0/grafana_metadata.py
```diff
@@ -0,0 +1,219 @@
...
+    """The provider side of the grafana-metadata relation."""
+
+    def __init__(
+        self,
+        relations: RelationMapping,
+        app: Application,
+        grafana_uid: str,
+        direct_url: AnyHttpUrl,
+        ingress_url: Optional[AnyHttpUrl] = None,
+        relation_name: str = GRAFANA_METADATA_RELATION_NAME,
```
simskij, line 180: Too many arguments. Please refactor or place them in a data class. This can wait until **after** merging though.

simskij (approved):

lib/charms/grafana_k8s/v0/grafana_metadata.py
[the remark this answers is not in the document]
simskij: We can address that in a separate issue. The implication isn't anything more than "it does not really make a lot of sense".

tests/integration/test_grafana_metadata.py
simskij: We should make that calculation part of the PR CI. I'll raise a separate issue for it.

PR #393, opened 2025-02-27
fix: fix the grafana oauth integration test

wood-push-melon:
This pull request aims to add the OAuth2 integration test back. The previous failure was due to something wrong with the headless chronium.

### DIFF
--- a/tests/integration/test_grafana_oauth.py
+++ b/tests/integration/test_grafana_oauth.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Tests the oauth library using the Canonical Identity Stack.
+
+It tests that the grafana charm can provide Single Sign-On Services to users
+with the oauth integration.
+"""
+
+import logging
+from pathlib import Path
+from helpers import oci_image
+
+import os
+import pytest
+import requests
+from playwright.async_api import Page, BrowserContext
+from pytest_operator.plugin import OpsTest
+
+from oauth_tools import (
+    deploy_identity_bundle,
+    get_reverse_proxy_app_url,
+    complete_auth_code_login,
+    access_application_login_page,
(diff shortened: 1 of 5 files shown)


tests/integration/test_grafana_oauth.py
```diff
@@ -0,0 +1,120 @@
...
+    self_signed_certificates_app_name: str,
+    ext_idp_service: ExternalIdpService,
+):
+    # Deploy identity bundle
+    await deploy_identity_bundle(
+        ops_test=ops_test, bundle_channel="0.2/edge", ext_idp_service=ext_idp_service
+    )
+
+    # Deploy grafana
+    await ops_test.model.deploy(  # pyright: ignore
```
lucabello, line 59: if the problem is that `ops_test.model` could be none, you can address it by adding the following at the top of the function:

```
    asserts ops_test.model
```

This allows to remove the `# pyright: ignore` comment, and should work for all the others as well

lucabello (changes requested):

charmcraft.yaml
```diff
@@ -116,7 +116,7 @@ platforms:
 parts:
   charm:
     charm-binary-python-packages: [cryptography, jsonschema, pydantic, pydantic-core, maturin]
-    build-packages: [git]
+    build-packages: [git, rustc, cargo]
```
ca-scribner, line 119: This was probably needed because the uv.lock file pointed to an older version of at least one of `[cryptography, jsonschema, pydantic, pydantic-core, maturin]`, thus the charm needed to build the binary itself.  This is the same in several of our charms, so maybe for this PR we should just update the lock file (which I think will fix things) and leave this change for discussing separately?

MichaelThamm (changes requested):
Looks good to me besides removing the pyright: ignore commnets

MichaelThamm:
LGTM, @sed-i could you give this a final review?

MichaelThamm (approved):

PR #397, opened 2025-03-03
feat: improve typing on GrafanaMetadataAppData, fix lib tests

ca-scribner:
Previously, the GrafanaMetadataAppData object looked to type checkers like it accepted AnyHttpUrl types even though it would also accept string types.  This led to people either needing to instantiate their own AnyHttpUrl objects or use type check ignore statements, neither of which are nice.

This change improves the typing of the model, allowing for str to be accepted but maintaining the full validation of an AnyHttpUrl type.

Also included here is a patch to the metadata lib test suite.  The previous test suite accidentally relies on a [bug in ops about deep copying](https://github.com/canonical/operator/issues/1605).  When that bug is fixed, the previous test suite would break.  The change here is the intended way of using the scenario Relation object


MichaelThamm (approved):
LGTM

lucabello:
@ca-scribner could you fix the static checks?

```python
grafana_metadata.publish(
    grafana_uid=self.unique_name,
    ingress_url=external_url,  # external_url is "str|None" instead of "AnyHttpUrl"
    direct_url=internal_url,  # internal_url is "str" instead of "AnyHttpUrl"
)
```

PR #398, opened 2025-03-03
feat: grafana_base_url in Grafana source

MichaelThamm:
## Issue
Loki needs to know the Grafana base URL for its `ruler` config which may have an ingress. This is needed to correctly render the Source button URL to Grafana when Loki alert rules are firing:
image
 
## Solution
Is this relation the best way to give Loki the ingress URL from Grafana? It makes sense since this scenario requires the `grafana-source` relation, otherwise Loki logs won't make it to Grafana and this scenario won't be possible.
- [x] Add tests for the library changes
- [x] Seems to collide with: https://github.com/canonical/grafana-k8s-operator/pull/388 likely not a good solution because:
	- > It is recommended that those charms also set `limit=1` for that relation in charmcraft.yaml

## Context
- https://github.com/canonical/alertmanager-k8s-operator/issues/293
- https://github.com/canonical/loki-k8s-operator/pull/494
- https://grafana.com/docs/loki/latest/configure/#ruler

## Testing Instructions
1. Create an always firing Loki alert rule in the [Zinc charm](https://github.com/jnsgruk/zinc-k8s-operator):
```yaml
alert: HostHighLogErrorRate
expr: count_over_time({juju_application="zinc"} [1h]) > -1
for: 0m
labels:
  severity: warning
annotations:
  summary: High error rate in logs (instance {{ $labels.instance }})
  description: "High error rate in logs\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
```
2. Pack the zinc charm
3. Deploy Zinc
4. Deploy cos-lite
```
curl -L https://raw.githubusercontent.com/canonical/cos-lite-bundle/main/overlays/offers-overlay.yaml -O
juju deploy cos-lite --trust --overlay ./offers-overlay.yaml
juju deploy 
```
5. Pack the Loki and Grafana charms with the changes in this PR and the tandem PR
6. `juju refresh` both charms with the packed charms
7. Relate Zinc to Loki for logging and generate some logs in the Zinc UI (log in or run a query)
8. Check the Alertmanager UI for the Source button link to Grafana ingress URL instead of Loki's ingress URL

## Upgrade Notes
The `get_source_uids` is left backwards compatible (since it is public) but should not be used anymore since `get_source_data` duplicates its contents and is more flexible for future additions to Grafana source data by updating the `GrafanaSourceData` dataclass.


tests/unit/test_datasources.py
```diff
@@ -123,3 +132,5 @@ def __init__(self, framework: Framework):
         charm = mgr.charm
         # THEN we can see no datasource uids via the provider
         assert not charm.source_provider.get_source_uids()  # type: ignore
+        # AND we can see no Grafana base URL via the provider
+        assert not charm.source_provider.get_grafana_base_urls()  # type: ignore
```
MichaelThamm, line 136: Do we want to care about this since this is a new public method?

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -446,9 +462,21 @@ def get_source_uids(self) -> Dict[str, Dict[str, str]]:
...
+            )
+            data[grafana_uid] = grafana_data
+        return data
 
-            uids[grafana_uid] = json.loads(app_databag.get("datasource_uids", "{}"))
-        return uids
+    def get_source_uids(self) -> Dict[str, Dict[str, str]]:
+        """Get the datasource UID(s) assigned by the remote end(s) to this datasource.
+
+        Use the `get_source_data` public method instead.
```
lucabello, line 475: You could add the word "deprecated" on this line.

lucabello (approved):

PR #400, opened 2025-03-04
feat: add rust build tools to fix dependency builds

ca-scribner:


ca-scribner:
This appears good to go.  The failure is because `main` has an unpublished library (unpublished because the previous merge failed, I think)

lucabello (approved):

PR #405, opened 2025-03-21
fix admin password desync

### LINKED ISSUE #262: admin password has been changed: error on get-admin-password
### Bug Description

```
$ juju run grafana/0 get-admin-password 
Running operation 5 with 1 task
  - task 6 on unit-grafana-0

Waiting for task 6...
admin-password: Admin password has been changed by an administrator
url: http://10.42.42.42/cos-grafana
```

### To Reproduce

deploy cos-lite --edge

### Environment

microk8s

### Relevant log output

```shell
n/a
```

### Additional context

hard to reproduce.

Known workaround:
`juju ssh --container grafana grafana/0 /usr/share/grafana/bin/grafana-cli --homepath "/usr/share/grafana" admin reset-admin-password foobarbaz`

PietroPasotti:
## Issue
Fixes #262 

## Solution
Replace storedstate with a Secret, use it to sync credentials between peers for the purposes of get-admin-password.

## Testing Instructions

```
jhack deploy 
j run grafana/0 get-admin-password
j add-unit grafana -n 2
j run grafana/1 get-admin-password  # should be the same password!
j run grafana/2 get-admin-password  # should be the same password!
```

UI should be accessible with the same credentials at all 3 urls.

## Upgrade Notes
Should be automagic.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,6 @@ import json
 import logging
 import os
 import re
-import secrets
 import socket
 import string
 import subprocess
@@ -35,7 +34,7 @@ from urllib.parse import urlparse
 
 import yaml
 from cosl import JujuTopology
-from ops import main, ModelError
+from ops import main
 from ops.charm import (
     ActionEvent,
     CharmBase,
@@ -46,7 +45,7 @@ from ops.charm import (
     RelationJoinedEvent,
     UpgradeCharmEvent,
 )
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, Port, SecretNotFoundError
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, Port
 from ops.pebble import (
     APIError,
     ConnectionError,
@@ -56,6 +55,7 @@ from ops.pebble import (
     PathError,
     ProtocolError,
 )
+from secret_storage import SecretStorage
 
 from charms.catalogue_k8s.v1.catalogue import CatalogueConsumer, CatalogueItem
 from charms.certificate_transfer_interface.v0.certificate_transfer import (
@@ -89,6 +89,7 @@ from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
 from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer, charm_tracing_config
 from charms.traefik_k8s.v0.traefik_route import TraefikRouteRequirer
 from grafana_client import Grafana, GrafanaCommError
+from secret_storage import generate_password
 
 logger = logging.getLogger()
 
@@ -159,6 +160,8 @@ class GrafanaCharm(CharmBase):
         self._grafana_config_ini_hash = None
         self._grafana_datasources_hash = None
         self._topology = JujuTopology.from_charm(self)
+        self._secret_storage = SecretStorage(self, "admin-password",
+                                             default=lambda: {"password": generate_password()})
 
         # -- cert_handler
         self.cert_handler = CertHandler(
@@ -1317,7 +1320,6 @@ class GrafanaCharm(CharmBase):
         datasources_string = yaml.dump(datasources_dict)
         return datasources_string
 
-
     class GetAdminPWDFailures:
         waiting_for_leader = "Still waiting for the leader to generate an admin password..."
         not_reachable = 'Grafana is not reachable yet. Please try again in a few minutes'
@@ -1328,6 +1330,7 @@ class GrafanaCharm(CharmBase):
     def _on_get_admin_password(self, event: ActionEvent) -> None:
         """Returns the grafana url and password for the admin user as an action response."""
         admin_password = self.admin_password
+
         if not self.unit.is_leader() and admin_password is None:
             return event.fail(self.GetAdminPWDFailures.waiting_for_leader)
 
@@ -1371,41 +1374,10 @@ class GrafanaCharm(CharmBase):
     @property
     def admin_password(self) -> Optional[str]:
         """The admin password."""
-        # check if secret exists already
-        secret = None
-        secret_label = "admin-password"
-        secret_content_key = "password"
-
-        try:
-            secret = self.model.get_secret(label=secret_label)
-        except SecretNotFoundError:
-            logger.info("admin-password secret does not exist yet")
-        except ModelError:
-            logger.exception("error retrieving admin-password secret")
-        except:
-            raise
-
-        # if we're leader and have already generated the secret in a previous run,
-        # or we're a follower and the leader has given us a secret already: fetch the password
-        if secret:
-            logger.debug("secret found: returning key")
-            return secret.get_content()[secret_content_key]
-
-        # if we're a leader: generate the password and drop it in a secret
-        if self.unit.is_leader():
-            logger.info("leader: generating secret")
-
-            password = generate_password()
-            self.app.add_secret(
-                {secret_content_key: password},
-                label=secret_label,
-                description="Secret to hold the admin password for this grafana instance."
-            )
-            return password
-
-        # if we're a follower and the leader hasn't generated a password and put it in a secret yet,
-        # then we return None as we can't know yet what the password is
-        return None
+        contents = self._secret_storage.contents
+        if not contents:
+            return None
+        return contents.get('password')
 
     def _poll_container(
             self, func: Callable[[], bool], timeout: float = 2.0, delay: float = 0.1
@@ -1708,14 +1680,6 @@ class GrafanaCharm(CharmBase):
             ingress_url=external_url,  # pyright: ignore
             direct_url=internal_url,  # pyright: ignore
         )
-
-
-def generate_password() -> str:
-    """Generates a random 12 character password."""
-    # Really limited by what can be passed into shell commands, since this all goes
-    # through subprocess. So much for complex password
-    chars = string.ascii_letters + string.digits
-    return "".join(secrets.choice(chars) for _ in range(12))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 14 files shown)


src/charm.py
```diff
@@ -1141,6 +1136,13 @@ def _build_layer(self) -> Layer:
                 }
             )
 
+        # If we're followers, we don't need to set any credentials on the grafana process.
+        # This Grafana instance will inherit them automatically from the replication primary (the leader).
```
sed-i, line 1140: Great find! Add a doc link?
I wonder if this was the root cause of the original issue.

src/charm.py
```diff
@@ -1318,62 +1318,93 @@ def _generate_datasource_config(self) -> str:
         return datasources_string
 
     def _on_get_admin_password(self, event: ActionEvent) -> None:
-        """Returns the grafana url and password for the admin user as an action response."""
+        """Returns the grafana url and password for the admin user as an action response.
+
+        This operation is leader-only.
```
sed-i, line 1323:
```suggestion
```

src/charm.py
```diff
@@ -1318,62 +1318,93 @@ def _generate_datasource_config(self) -> str:
...
             return
 
         if pw_changed:
+            if self.unit.is_leader():
+                msg = "Admin password has been changed by an administrator."
+            else:
+                # it takes a little bit of time for grafana to settle on the
+                # authentication data provided by the leader unit
+                msg = ("Admin password may have been changed by an administrator. "
+                       "To be sure, run this action on the leader unit.")
```
sed-i, line 1354: Nice!

src/charm.py
```diff
@@ -1318,62 +1318,93 @@ def _generate_datasource_config(self) -> str:
...
-                cast(str, self.model.config["admin_user"]), self._get_admin_password()
+                cast(str, self.model.config["admin_user"]), admin_password
             )
         except GrafanaCommError as e:
             event.fail(f"Grafana is not reachable yet: {e}. Please try again in a few minutes.")
             return
 
         if pw_changed:
+            if self.unit.is_leader():
+                msg = "Admin password has been changed by an administrator."
```
sed-i, line 1349: Can we add instructions to resolve? I guess it would be:
- Run the grafana cli to set admin password.
- Use the same value to update the secret.

BTW should we observe secret changed for manual password rotation?

src/charm.py
```diff
@@ -1318,62 +1318,93 @@ def _generate_datasource_config(self) -> str:
...
-    def _generate_admin_password(self) -> None:
-        """Generate the admin password if it's not already in stored state, and store it there."""
-        if not self._stored.admin_password:  # type: ignore[truthy-function]
-            logger.debug("Grafana admin password is not in stored state, so generating a new one.")
-            self._stored.admin_password = self._generate_password()
-
-    def _get_admin_password(self) -> str:
-        """Returns the password for the admin user.
+    @property
+    def admin_password(self) -> Optional[str]:
```
sed-i, line 1368: Would we benefit from a special kind of a "secret manager" lib? A lib that is only addressing "secrets scoped to the charm itself". It seems like it's going to be a recurring pattern.
Would be interesting to compare to how tls-certificates does it for the private key.

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+        yield
+
+
+@pytest.fixture
+def ctx():
+    return testing.Context(GrafanaCharm)
+
+
+def check_valid_password(s: str):
+    assert re.match(r"[A-Za-z0-9]{12}", s)
```
sed-i, line 49: Password vocabulary should be a separate test.

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+
+def test_can_get_password(ctx):
+    # GIVEN a grafana leader unit
+    state = testing.State(leader=True)
+
+    # WHEN we receive any hook
+    with ctx(ctx.on.update_status(), state) as mgr:
+        # THEN the .admin_password attribute returns a valid password
+        pwd = mgr.charm.admin_password
+        check_valid_password(pwd)
```
sed-i, line 60: Per interface, `def admin_password(self) -> Optional[str]:` and test scope, how about:
```suggestion
        # THEN the .admin_password attribute returns a string with length > 0
        pwd = mgr.charm.admin_password
        assert isinstance(pwd, str)
        assert len(pwd) > 0
```

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+    pwd = "abcde"
+    state = testing.State(leader=leader, secrets={
+        testing.Secret(tracked_content={"password": pwd}, label="admin-password")})
+
+    # WHEN we run the get-admin-password action
+    with grafana_ready(True):
+        with password_changed(False):
+            ctx.run(ctx.on.action('get-admin-password'), state)
+
+    # THEN we obtain the expected result
```
sed-i, line 81:
```suggestion
    # THEN the secret's value matches the pre-existing secret.
```

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+    state = testing.State()
+
+    # WHEN we run the get-admin-password action
+    with password_changed(False):
+        with grafana_ready(True):
+            with pytest.raises(testing.ActionFailed) as failure:
+                ctx.run(ctx.on.action('get-admin-password'), state)
+
+    # THEN the action fails with this message
+    assert failure.value.message=='Still waiting for the leader to generate an admin password...'
```
sed-i, line 96: I think the string literal has no public interface significance, so I'd make this assertion less brittle with something like:

```suggestion
    assert len(failure.value.message) > 0
```

But wdyt?

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+
+@pytest.mark.parametrize("leader", (True, False))
+def test_action_grafana_down(ctx, leader):
+    # GIVEN a grafana unit, leader or not, with the secret ready
+    pwd = "abcde"
+    state = testing.State(leader=leader, secrets={
+        testing.Secret(tracked_content={"password": pwd}, label="admin-password")})
+
+    # WHEN we run the get-admin-password action
+    # BUT grafana is not ready
```
sed-i, line 108:
```suggestion
    # AND GIVEN grafana is not ready
    # WHEN we run the get-admin-password action
```

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+    with grafana_ready(False):
+        with pytest.raises(testing.ActionFailed) as failure:
+            ctx.run(ctx.on.action('get-admin-password'), state)
+
+    # THEN the action fails with this message
+    assert failure.value.message=='Grafana is not reachable yet. Please try again in a few minutes'
```
sed-i, line 114: Similar comment as above about brittleness.
In a past life, string messages were all stored as constants so were reusable in tests. Not sure if we want to explore such practice.

tests/scenario/test_charm.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@pytest.mark.parametrize("leader", (True, False))
+def test_action_password_changed(ctx, leader):
+    # GIVEN a grafana unit with the secret created already
+    pwd = "abcde"
+    state = testing.State(leader=leader, secrets={
+        testing.Secret(tracked_content={"password": pwd}, label="admin-password")})
+
+    # WHEN we run the get-admin-password action
```
sed-i, line 125:
```suggestion
    # AND GIVEN the admin password was changed behind the scenes
    # WHEN we run the get-admin-password action
```

tests/scenario/test_charm.py
sed-i, line 1: Should all the tests below be under a `test_admin_password.py` file? I know we've been dumping everything into `test_charm` so far but it seems like a practice we should correct.

src/charm.py
PietroPasotti: couldn't find any docs on that unfortunately. the only evidence I have is empirical :/

src/charm.py
PietroPasotti: I'm not sure what user story this is covering.
1. user deploys charm
2. charm generates random pwd
3. user shells in and manually changes the password to something else

at this point, why would the user ask again the charm 'what is my password'? you just changed it!
and, why would the user want to let the charm know 'this is my new password'? so that you can ask the charm 'what was it again?'

if the user forgets the password, then they should repeat step 3.
That, or we offer a juju action to 'reset' the pwd i.e. repeat step 2.

Either way this is out of scope for this PR.
Opening a new issue to track.

tests/scenario/test_charm.py
PietroPasotti: I think in the future we may add more failure modes, so I think it's not enough to verify that the action failed, we want to assert that it failed for the reason we expect it to.
It's the same reason why if possible you should prefer `assertRaises(MySpecificError)` over `assertRaises(Exception)`.

I get that this is brittle though, perhaps we could lift the error message to a global (or charm class attr) and use it from both the charm and the tests, so if it changes, it only has to change in a single place?

tests/scenario/test_charm.py
PietroPasotti: hah yes that's what I was suggesting above as workaround. will do that.

src/charm.py
PietroPasotti: I was thinking about that too, but it's not something for this PR imho.

tests/scenario/test_charm.py
PietroPasotti: done; check it out

src/charm.py
sed-i: This is the first time we introduce this practice for secrets, and it is likely to spread elsewhere. If not a charmlib then perhaps a separate .py file next to charm.py?

PietroPasotti: done

sed-i (approved):
Great job, thanks so much @PietroPasotti!

PietroPasotti:
I'll leave it to @lucabello to sort the CI out :rainbow:

PR #407, opened 2025-03-26
tests: get the output relation from the output state

tonyandrewmeyer:
## Issue

A few tests are asserting output data against the `Relation` object used in the input state. This works in the current version of ops-scenario, but that's because of a bug, which is fixed in the upcoming version.

## Solution

Adjust the tests to get the `Relation` from the output state.

## Context

N/A

## Testing Instructions

Run the unit tests with ops 2.19 and with canonical/operator@main.

## Upgrade Notes

N/A


dimaqq (approved):

lucabello:
Hey @tonyandrewmeyer, thanks for the PR! It seems like this was already addressed and merged to main separately, so I'll close this.

Thanks for the contribution though!

PR #421, opened 2025-05-19
Refactor charm to use the `reconcile` pattern

michaeldmitry:
This PR refactors the charm codebase to follow the `reconciler` pattern.

## Context
- The `reconcile` flavor attempted in this PR assumes a separate file for each workload managed by this charm (i.e. `grafana.py` and `litestream.py`) where each workload runs its own reconcile logic. It should be relatively easy to go from this favor to any other flavor of `reconcile`. 
- We still need to observe some events due to how some of the libs work.

## Drive-bys
- Migrate most of the `harness` tests to `scenario`


src/litestream.py
```diff
@@ -0,0 +1,114 @@
...
+
+        return layer
+
+
+
+    def reconcile(self):
+        """Unconditional control logic."""
+        if self._container.can_connect():
+            if self._reconcile_config():
+                self.restart_litestream()
```
simskij, line 71:
```suggestion
        if not self._container.can_connect():
        	return
        if not self._reconcile_config():
          return
        self.restart_litestream()
```

simskij: In addition, maybe `_reconcile_config` should not return a boolean? Based on the name it seems like this is an unexpected side-effect.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+                oauth_config: Optional[OauthProviderConfig] = None,
+                enable_profiling: bool = False,
+                enable_reporting: bool = True,
+                enable_external_db:  bool = False,
+                admin_user: Optional[str] = None,
+                admin_password: Optional[str] = None,
+                tls_config: Optional[TLSConfig] = None,
+                trusted_ca_certs: Optional[str] = None,
+                dashboards: List[Dict] = [],
+                provision_own_dashboard: bool = False,
```
simskij, line 75: I would encapsulate these in a `GrafanaConfig` object, but keep `container`, `is_leader`, and `peers` as they are charm-level concerns.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+        version_output, _ = self._container.exec(["grafana-server", "-v"]).wait_output()
+        # Output looks like this:
+        # Version 8.2.6 (commit: d2cccfe, branch: HEAD)
+        result = re.search(r"Version (\d*\.\d*\.\d*)", version_output)
+        if not result:
+            return ""
+        return result.group(1)
+
+
+    def _get_hash_for_file(self, file: str) -> str:
```
simskij, line 120: This should probably be moved somewhere else in the file as it's currently injected between two `@property`s

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+            if any(
+                    (self._reconcile_tls_config(),
+                    self._reconcile_trusted_ca(),
+                    self._reconcile_config(),
+                    self._reconcile_ds_config(),
+                    self._reconcile_dashboards_config(),
+                    self._reconcile_pebble_plan(),
+                    )
+                   ):
+                self._restart_grafana()
```
simskij, line 321: Same comment as elsewhere, it is not really clear from the called methods that they also indicate whether we should restart the workload.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+        if not db_config:
+            return ""
+
+        db_url = "{0}://{1}:{2}@{3}/{4}".format(
+            db_type,
+            db_config.get("user"),
+            db_config.get("password"),
+            db_config.get("host"),
+            db_config.get("name"),
+        )
```
simskij, line 702: An f-string would improve the readability here.

simskij (approved):
None of my comments are blocking.

src/litestream.py
PietroPasotti: maybe a more robust pattern is to do something like:

``` 
changes = []
_reconcile_smh(changes)
_reconcile_smh_else(changes)
if any(changes):
    _restart()

```

src/grafana.py
PietroPasotti: not sure I agree with that. That means worse DX as now you'd have to import a new type which is in practice just a namespace.
Perhaps we could encapsulate the grafana config generation in a new object (which'd mean having not only these attributes in there, but also methods to generate the config, etc...)

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+
+
+class GrafanaCommError(Exception):
+    """Raised when comm fails unexpectedly."""
+
+
+class Grafana:
+    """Grafana workload."""
+
+    def __init__(self,
```
sed-i, line 57: Just pointing out that the long list of unrelated constructor args here is a code smell, likely because a "workload" class is the wrong abstraction. Feel free to open an issue if you have a further refactoring idea after this exercise.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+        if self._enable_external_db:
+            peer_data = self._peers.get_peer_data("database")
+            if not peer_data:
+                return None
+            return peer_data
+        return None
+
+    def reconcile(self):
+        """Unconditional control logic."""
+        if self._container.can_connect():
```
sed-i, line 307: Not sure if in this particular case we want to silently absorb a can_connect.
Seems like grafana.reconcile() should fail loudly, and rely on the caller to have a can_connect guard, but perhaps I'm missing some context.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+                    self._reconcile_config(),
+                    self._reconcile_ds_config(),
+                    self._reconcile_dashboards_config(),
+                    self._reconcile_pebble_plan(),
+                    )
+                   ):
+                self._restart_grafana()
+
+
+    def _reconcile_provisioning_dirs(self):
```
sed-i, line 324: "reconcile" here seems out of place. Should it just be "provision_dirs()"?

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+            self._container.exec(["update-ca-certificates", "--fresh"]).wait()
+        else:
+            if self._container.exists(TRUSTED_CA_CERT_PATH):
+                any_change = True
+                self._container.remove_path(TRUSTED_CA_CERT_PATH, recursive=True)
+
+        return any_change
+
+
+    def _reconcile_tls_config(self) -> bool:
```
sed-i, line 419: Similar to prev comment, I think the term "reconcile" is being overused here throughout.

src/grafana.py
```diff
@@ -0,0 +1,797 @@
...
+    @property
+    def is_ready(self) -> bool:
+        """Checks whether the Grafana server is up and running yet.
+
+        Returns:
+            :bool: indicating whether the server is ready
+        """
+        return True if self.build_info.get("database", None) == "ok" else False
+
+    def password_has_been_changed(self, username: str, passwd: str) -> bool:
```
sed-i, line 750: I think keeping these in a separate, thin "grafana_client" lib is neater.
With this change, we need to instantiate this whole object just to do "is_ready" on an endpoint.

src/peer.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2025 Canonical
+# See LICENSE file for licensing details.
+"""Peer relation class."""
+import json
+from typing import Any, Optional
+from ops import Application, Relation
+
+class Peer:
+    """A helper class to manage peer relation data."""
```
sed-i, line 9: At first glance it seems this class is applicable to any relation, not just peer relation..?

tests/integration/test_trusted_certificates.py
```diff
@@ -69,7 +68,7 @@ async def test_certs_created(ops_test: OpsTest):
     relation_cert = stdout.rstrip()
 
     # Get pushed cert
-    received_cert_path = f"/usr/local/share/ca-certificates/trusted-ca-cert-{relation_id}-ca.crt"
+    received_cert_path = "/usr/local/share/ca-certificates/trusted-ca-cert.crt"
```
sed-i, line 71: Dropping the rel-id from the path relies on `limit: 1`, which is not something we'd want to assume going forward. What prompted this change?

tests/unit/test_config_reporting_enabled.py
```diff
@@ -1,16 +1,10 @@
...
-    Container(name="grafana", can_connect=True),
-    Container(name="litestream", can_connect=True),
-]
-
-
-def test_reporting_enabled(ctx):
+def test_reporting_enabled(ctx, base_state):
     # GIVEN the "reporting_enabled" config option is set to True
-    state = State(leader=True, config={"reporting_enabled": True}, containers=containers)
+    state = replace(base_state, config={"reporting_enabled": True})
```
sed-i, line 7: This is probably too DRY for a test. I'd opt for a more "DAMP" approach here.

tests/unit/test_external_url.py
```diff
@@ -2,166 +2,118 @@
...
-
-        # Peer relation
-        self.app_name = "grafana-k8s"
-        self.peer_rel_id = self.harness.add_relation("grafana", self.app_name)
-
-        # Auth relation
-        # FIXME why tests fail when this relation is not created?
-        self.grafana_auth_rel_id = self.harness.add_relation("grafana-auth", "auth_provider")
-
-        self.harness.begin_with_initial_hooks()
```
sed-i, line 63: Note that this change replaces "begin with initial hooks" with "WHEN any event is fired".
1. The comment `WHEN any event is fired` is misleading: we're only doing "update-status" currently.
2. "begin with initial hooks" gives more confidence that charm code operates correctly, than firing an orphan "update-status".

src/litestream.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if self._is_leader:
+            self._peers.set_peer_data("replica_primary", socket.gethostbyname(self._fqdn))
+            config["LITESTREAM_ADDR"] = "{}:{}".format(
+                socket.gethostbyname(self._fqdn), "9876"
+            )
+        else:
+            config["LITESTREAM_UPSTREAM_URL"] = "{}:{}".format(
+                self._peers.get_peer_data("replica_primary"), "9876"
+            )
```
Abuelodelanada, line 43:
```suggestion
            config["LITESTREAM_UPSTREAM_URL"] = f"{self._peers.get_peer_data('replica_primary')}:{9876}"
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1676,14 +1676,22 @@ def _set_default_data(self) -> None:
 
     def set_peer_data(self, key: str, data: Any) -> None:
         """Put information into the peer data bucket instead of `StoredState`."""
-        self._charm.peers.data[self._charm.app][key] = json.dumps(data)  # type: ignore[attr-defined]
+        peers = self._charm.peers  # type: ignore[attr-defined]
+        if not peers or not peers.data:
+            logger.info("set_peer_data: no peer relation. Is the charm being installed/removed?")
+            return
```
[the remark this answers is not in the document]
michaeldmitry: Uhm, no we're not guarding at call site. Could you please point out where we're doing that?
+ This is the grafana side, so it's safe to assume the impact is contained. Do you agree?

src/charm.py
```diff
@@ -88,44 +63,15 @@
 from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
 from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer, charm_tracing_config
 from charms.traefik_k8s.v0.traefik_route import TraefikRouteRequirer
-from grafana_client import Grafana, GrafanaCommError
+from grafana import Grafana, GrafanaCommError
 from secret_storage import generate_password
+from litestream import Litestream
+from peer import Peer
+from models import DatasourceConfig, PebbleEnvConfig, TLSConfig, TracingConfig
+from constants import PEER_RELATION, WORKLOAD_PORT, OAUTH_SCOPES, CA_CERT_PATH, GRAFANA_WORKLOAD, DATABASE_RELATION, PROFILING_PORT, OAUTH_GRANT_TYPES, REQUIRED_DATABASE_FIELDS, VALID_AUTHENTICATION_MODES
```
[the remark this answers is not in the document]
michaeldmitry: Made it a multi-line import

src/charm.py
```diff
@@ -152,17 +98,13 @@ def __init__(self, *args):
...
         # -- initialize states --
-        self.name = "grafana"
-        self.containers = {
-            "workload": self.unit.get_container(self.name),
-            "replication": self.unit.get_container("litestream"),
-        }
-        self._grafana_config_ini_hash = None
-        self._grafana_datasources_hash = None
         self._topology = JujuTopology.from_charm(self)
+        self._fqdn = socket.getfqdn()
```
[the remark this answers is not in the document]
michaeldmitry: It is used somewhere else. I just forgot to use it :)

src/charm.py
```diff
@@ -178,14 +120,6 @@ def __init__(self, *args):
         # TraefikRouteRequirer expects an existing relation to be passed as part of the constructor,
         # so this may be none. Rely on `self.ingress.is_ready` later to check
         self.ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
-        self.framework.observe(self.on["ingress"].relation_joined, self._configure_ingress)
-        self.framework.observe(self.ingress.on.ready, self._on_ingress_ready)  # pyright: ignore
-        self.framework.observe(self.on.leader_elected, self._configure_ingress)
-        self.framework.observe(self.on.config_changed, self._configure_ingress)
-        self.framework.observe(self.cert_handler.on.cert_changed, self._configure_ingress)
```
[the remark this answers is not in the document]
michaeldmitry: Yup

src/charm.py
```diff
@@ -152,17 +98,13 @@ def __init__(self, *args):
...
-        self.name = "grafana"
-        self.containers = {
-            "workload": self.unit.get_container(self.name),
-            "replication": self.unit.get_container("litestream"),
-        }
-        self._grafana_config_ini_hash = None
-        self._grafana_datasources_hash = None
         self._topology = JujuTopology.from_charm(self)
+        self._fqdn = socket.getfqdn()
+        self.peers = Peer(app=self.app, peers=self.model.get_relation(PEER_RELATION))
```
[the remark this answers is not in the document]
michaeldmitry: We need the instance attribute to be `peers` because apparently the charm libs reference `self._charm.peers`.
I think I'll change the class name to `PeerData`. That way it's less confusing and is a singular noun.

src/charm.py
```diff
@@ -207,141 +141,205 @@ def __init__(self, *args):
...
-        )
-        self.framework.observe(
-            self.workload_tracing.on.endpoint_removed,  # type: ignore
-            self._on_workload_tracing_endpoint_removed,
+            self.on.get_admin_password_action,  # pyright: ignore
+            self._on_get_admin_password,
         )
 
-        # -- cert_handler observations
+        # FIXME: we need to observe these events as they contain the required data
```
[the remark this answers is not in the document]
michaeldmitry: ideally we should fix the libs at some point to work with the reconcile approach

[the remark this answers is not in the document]
michaeldmitry: Updated the comment to convey an action to update the libs

src/constants.py
```diff
@@ -0,0 +1,27 @@
...
+OAUTH_GRANT_TYPES = ["authorization_code", "refresh_token"]
+GRAFANA_WORKLOAD = "grafana"
+VALID_AUTHENTICATION_MODES = {"proxy"}
+CONFIG_PATH = "/etc/grafana/grafana-config.ini"
+PROVISIONING_PATH = "/etc/grafana/provisioning"
+DATASOURCES_PATH = "/etc/grafana/provisioning/datasources/datasources.yaml"
+DASHBOARDS_DIR = f"{PROVISIONING_PATH}/dashboards"
+GRAFANA_CRT_PATH = "/etc/grafana/grafana.crt"
+GRAFANA_KEY_PATH = "/etc/grafana/grafana.key"
+CA_CERT_PATH = "/usr/local/share/ca-certificates/cos-ca.crt"
```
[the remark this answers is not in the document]
michaeldmitry: I already did that by combining all certs received over `certificate_transfer` in one file and copying to `TRUSTED_CA_CERT_PATH`

Check `_trusted_ca_certs`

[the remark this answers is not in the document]
michaeldmitry: I removed `limit: 1` while we're at it

src/peer.py
michaeldmitry: uhmm not really

tests/unit/test_config_reporting_enabled.py
michaeldmitry: WDYM?

tests/unit/test_external_url.py
michaeldmitry: Added a bunch of events to run the test on
The idea is that it doesn't matter which event we trigger so I just put a bunch of "begin with initial hooks" events

src/charm.py
```diff
@@ -207,141 +141,205 @@ def __init__(self, *args):
...
 
-        # oauth relation observations
-        self.framework.observe(
-            self.oauth.on.oauth_info_changed,
-            self._on_oauth_info_changed,  # pyright: ignore
-        )
-        self.framework.observe(
-            self.oauth.on.oauth_info_removed,
-            self._on_oauth_info_changed,  # pyright: ignore
+        self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
```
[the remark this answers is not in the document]
sed-i: That's fine. Note that generally this may require relying on stored state (see pull/push status [here](https://discourse.charmhub.io/t/its-probably-ok-for-a-unit-to-go-into-error-state/13022#a-compromise-9)).

Probably not for this PR, but generally, if we really wanted, we could use the prioritization logic [directly](https://github.com/canonical/operator/blob/160c7e05cf74f6bcf7dfebd679a79a23d725b6e6/ops/model.py#L2087):
```python
collected_statuses = []
collected_statuses.append(...)
# ...
self.unit.status = model.StatusBase._get_highest_priority(collected_statuses)
```

src/grafana.py
sed-i: I meant that the can-connect guard should probably be in `__init__` before we call reconcile().
I.e. if someone is calling reconcile(), then they expect it to do its thing, and not quietly return after doing nothing.

src/peer.py
sed-i: Can you elaborate?
It seems like if we drop "peer" from this file, it would equally work for any relation, because all we do here is get data, check if there are other units on the relation, set data. How is this specific to peer relation?

tests/unit/test_config_reporting_enabled.py
sed-i: https://discourse.charmhub.io/t/best-practices-for-testing-charms/15172#p-33192-a-test-should-be-obvious-upon-inspection-damp-7

I wouldn't use `replace`, I think.

tests/unit/test_external_url.py
sed-i: It depends on how well juju behavior is reflected by the testing framework.
For example, at early lifecycle hooks, a peer relation doesn't exist yet, so if charm code assumes it exists, then "begin with initial hooks" would surface it.

Iiuc, parameterization with scenario does not give us the extra coverage harness gave us with "begin with initial hooks". Specifically, when we fire `install()`, do we have a peer relation in place? We shouldn't.

sed-i (approved):

src/peer.py
michaeldmitry: updated it to a generic relation class

PR #427, opened 2025-07-04
feat: Grafana source unit data flag

PietroPasotti:
Related charms which need this feature:
- https://github.com/canonical/mimir-coordinator-k8s-operator/pull/161
- https://github.com/canonical/loki-coordinator-k8s-operator/pull/77
- https://github.com/canonical/tempo-operators/pull/135

> [!NOTE]
> Non-coordinated-worker charms do not need this feature since [this existing code](https://github.com/canonical/grafana-k8s-operator/blob/1ee19d9d5a782cf9f68e911aa02041bd7ee5e538/lib/charms/grafana_k8s/v0/grafana_source.py#L759) removes their datasources already

Allows charms to disable setting `grafana_source_host` in unit data.

Issue is: ATM tempo and other coordinated workers have no way to share a single datasource endpoint once they are scaled up.

As a result all units end up with a grafana-source-host field in their unit databags:

image

This PR adds a `is_ingress_per_app` boolean flag, defaulting to False to keep backwards compatibility.
A charm can pass True to tell the library **not** to put a datasource URL in the unit databag, therefore in grafana only a single datasource (pointing to the leader unit's FQDN/ingress address) will show up.

# Lib changes
## GrafanaSourceProvider
API:
```python
self.grafana_source = GrafanaSourceProvider(
    self,
    # snip ...
    refresh_event=[
        self.ingress.on.ready,
        self.ingress.on.revoked,
        # snip ... other events which require datasource reconcilation
    ],
    is_ingress_per_app=self.ingress.is_ready(),
)
```
## GrafanaSourceConsumer
Grafana compares old sources to new sources and buids the `sources_to_delete` list for datasource removal. This is in Grafana's `peer` databag which is used as the source of truth for the state of the datastore.
```json
{
  "sources": "{\"12\": [{\"unit\": \"mimir/1\", \"source_name\": \"juju_test-charm-ha-scaled-x3so_30c1619c-fba3-4e18-8b97-34c02d87090f_mimir_1\", \"source_type\": \"prometheus\", \"url\": \"http://192.168.88.12/test-charm-ha-scaled-x3so-mimir/prometheus\"}]}",
  "sources_to_delete": "[\"juju_test-charm-ha-scaled-x3so_30c1619c-fba3-4e18-8b97-34c02d87090f_mimir_2\", \"juju_test-charm-ha-scaled-x3so_30c1619c-fba3-4e18-8b97-34c02d87090f_mimir_0\"]"
}
```

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -435,6 +435,7 @@ class GrafanaSourceProvider(Object):
         self.framework.observe(events.relation_joined, self._set_sources_from_event)
         self.framework.observe(events.relation_changed, self._set_sources_from_event)
         self.framework.observe(events.relation_departed, self._set_sources_from_event)
+        self.framework.observe(events.relation_broken, self._set_sources_from_event)
         for ev in refresh_event:
             self.framework.observe(ev, self._set_unit_details)
 
@@ -638,7 +639,6 @@ class GrafanaSourceConsumer(Object):
                 else []
             )
             new_source_names = [something["source_name"] for something in new_sources[rel_id]]
-            # Get strings in old_source_names that are not in new_source_names
             new_to_delete_for_rel = [
                 name for name in old_source_names if name not in new_source_names
             ]


lucabello:
Note that presenting multiple datasources when scaled up is the desired behavior IF the coordinators are not using ingress. A blanket leader guard doesn't necessarily help; we need to either:
1. conditionally instantiate the `GrafanaDashboardProvider` class:
	```python
	if self.external_url and self.unit.is_leader():
		self.grafana_dashboards = GrafanaDashboardProvider(...)
	else:
		self.grafana_dashboards = GrafanaDashboardProvider(...)
	```
2. have a flag in `GrafanaDashboardProvider` to toggle whether all units are datasources:
	```python
	self.grafana_dashboards = GrafanaDashboardProvider(
		...,
		units_are_datasources=(self.external_url and self.unit.is_leader()),
	)
	```

I'm fine with 2, but I would name the flag after the higher-level effect (whether all units should be used as datasources) instead of naming it after the direct effect (unit data not being set), as it doesn't require a library user to reason about the implications of that.

PietroPasotti:
well the thing is, the high-level effect is not 'units-are-datasources', because also the leader is a unit and the caller is only able to control _this unit_.

The higher-level effect is 'is-_this_-unit-a-datasource'.
```
self.grafana_dashboards = GrafanaDashboardProvider(
	...,
	this_unit_is_datasource=(self.external_url and self.unit.is_leader()),
)
```
WDYT?

PietroPasotti:
Re the thought: we can make it public later if needed, for now I vote to keep it private.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -421,9 +425,16 @@ def __init__(
...
         self._source_port = source_port
+
+        # If there's no ingress, then each unit is a datasource.
+        # If there is an ingress, then only the leader is a datasource.
+        self._this_unit_is_datasource = (not is_ingress_per_app) or charm.unit.is_leader()
+
         self._source_url = self._sanitize_source_url(source_url)
 
         self.framework.observe(events.relation_joined, self._set_sources_from_event)
+        self.framework.observe(events.relation_changed, self._set_sources_from_event)
```
MichaelThamm, line 436: There may be a better way to register this, but without the `relation_changed` registration, the ingress context remains in the `grafana_source_host` keys even when there is no ingress relation. I added `relation_departed` as well for safety.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -595,10 +610,42 @@ def _on_grafana_source_relation_changed(self, event: Optional[CharmEvents] = Non
                 if source:
                     sources[rel.id] = source
 
+            to_delete = self._sources_to_delete(sources)
             self.set_peer_data("sources", sources)
-
+            self.set_peer_data("sources_to_delete", to_delete)
         self.on.sources_changed.emit()  # pyright: ignore
 
+    def _sources_to_delete(self, new_sources: Dict) -> List:
```
MichaelThamm, line 618: I wonder if this new logic now duplicates [the `GrafanaSourceConsumer::_remove_source_from_datastore` functionality](https://github.com/canonical/grafana-k8s-operator/blob/a2770608526f2c82ddc3d225870f1a48eaa87179/lib/charms/grafana_k8s/v0/grafana_source.py#L696). This was necessary prior to this PR to ensure that if, e.g.

1. Alertmanager is related to Grafana as a source provider at a scale > 1.
2. Then scales down and that unit's datasources need to be scheduled for deletion.

Now we compare the expected `sources` in peer data to the old which should cover all cases assuming we have registered the correct events.

Another issue that might arise is that we never clear the sources to delete until the next time it is overwritten. This should be fine, but it might be confusing for someone to check the peer data for `sources_to_delete` and see that there are sources in there although they are not in the Grafana datastore.

lib/charms/grafana_k8s/v0/grafana_source.py
PietroPasotti: I can't think of a better way. I'd also add relation-broken just to be sure. Can't hurt

Issue #428, opened 2025-07-04
tox -e fmt is bork


PietroPasotti:
solution: replace `ruff check --fix-only` with `ruff format`

PR #430, opened 2025-07-11
fix: use `ruff format` instead of `ruff check --fix-only` in `tox -e fmt`

PietroPasotti:
demo PR for what happens if we use `ruff format` instead of `ruff check --fix-only` in `tox -e fmt`


lucabello:
Since this wasn't addressed prior, closing this as it's already tracked in an issue #428

PR #432, opened 2025-07-17
Make the probe work for the live status

wilkmar:
## Issue
I noticed that the `probes/relation_dashboard_uid.py` currently fails when run against a live status/model.

## Solution
I am proposing this change that works fine with both:
- show-unit output file
- live model

Please have a look and let me know if it needs any improvements.


probes/relation_dashboard_uid.py
```diff
@@ -3,33 +3,39 @@
...
-            dashboards = json.loads(app_data["dashboards"])
-            for id in dashboards:
-                for meta in dashboards[id]:
-                    decoded_uid = _uid_from_encoded_dashboard(meta["content"])
-                    # 5. Check if the top-level UID exist and is not empty
-                    if not _is_valid_format(decoded_uid):
-                        raise Exception(
-                            f"Invalid dashboard UID ({id}) for relation-id ({relation_id})"
-                        )
+    for parent in juju_show_units.values():
```
MichaelThamm, line 10: I don't think "parent" gives the correct impression. Technically we are iterating through all the values of a Dict of `show_unit`s.
```suggestion
    for show_unit in juju_show_units.values():
```

wilkmar: Hey @MichaelThamm, I agree, the 'parent' name here is a bit unfortunate, however, I couldn't come up with a better  name. We have to consider to cases here:
1. the probe is ran against `juju show-unit` output in the yaml format. In such case the 'parent' element in the dictionary is the name of yaml file, ie:
```
 juju-doctor check --show-unit grafana-0-show-unit.yaml --probe file:///home/mwilk/devel/grafana-k8s-operator/probes/relation_dashboard_uid.py --verbose
> /tmp/tmp1aylw2v5/probes/_home_mwilk_devel_grafana-k8s-operator_probes_relation_dashboard_uid.py(12)show_unit()
-> for parent in juju_show_units.values():
(Pdb) l
  7
  8
  9  	def show_unit(juju_show_units):
 10  	    import pdb
 11  	    pdb.set_trace()
 12  ->	    for parent in juju_show_units.values():
 13  	        # iterate over all units if checking live status
 14  	        for unit_name, unit in parent.items():
 15  	            # check applies to grafana unit(s) only
 16  	            if GRAFANA in unit_name:
 17  	                # 1. Extract the relation-info
(Pdb) p juju_show_units
{'grafana-0-show-unit.yaml': {'grafana/0': {'workload-version': '9.5.3', ...
```

2. the probe is ran against live model, then the 'parent' element is the model's name:
```
juju-doctor check --model coslite --probe file:///home/mwilk/devel/grafana-k8s-operator/probes/relation_dashboard_uid.py --verbose
> /tmp/tmp3nzy2gzt/probes/_home_mwilk_devel_grafana-k8s-operator_probes_relation_dashboard_uid.py(12)show_unit()
-> for parent in juju_show_units.values():
(Pdb) l
  7
  8
  9  	def show_unit(juju_show_units):
 10  	    import pdb
 11  	    pdb.set_trace()
 12  ->	    for parent in juju_show_units.values():
 13  	        # iterate over all units if checking live status
 14  	        for unit_name, unit in parent.items():
 15  	            # check applies to grafana unit(s) only
 16  	            if GRAFANA in unit_name:
 17  	                # 1. Extract the relation-info
(Pdb) p juju_show_units
{'coslite': {'alertmanager/0': {'workload-version': '0.27.0', 'opened-ports': [], ...
```

And if there is multiple `--model` options added to the command, the `juju_show_units` dic will have all of them as 'parents':
```
juju-doctor check --model coslite --model cosha --probe file:///home/mwilk/devel/grafana-k8s-operator/probes/relation_dashboard_uid.py --verbose
> /tmp/tmpnzy8o2tf/probes/_home_mwilk_devel_grafana-k8s-operator_probes_relation_dashboard_uid.py(12)show_unit()
-> for parent in juju_show_units.values():
(Pdb) l
  7
  8
  9  	def show_unit(juju_show_units):
 10  	    import pdb
 11  	    pdb.set_trace()
 12  ->	    for parent in juju_show_units.values():
 13  	        # iterate over all units if checking live status
 14  	        for unit_name, unit in parent.items():
 15  	            # check applies to grafana unit(s) only
 16  	            if GRAFANA in unit_name:
 17  	                # 1. Extract the relation-info
(Pdb) p juju_show_units
{'coslite': {'alertmanager/0': {'workload-version': '0.27.0', 'opened-ports': [],
....
'cosha': {'alertmanager/0': {'workload-version': '0.27.0', 'opened-ports': [],...
```

Please let me know what you think. I will also submit an updated commit where I identify grafana units based on a unit's charm, not the unit name (which is the case now).

MichaelThamm: You have more time spent on this probe than me at this point, so I am happy to take your advice. I do think `parent` is too generic in either case: live model or static file. Since `juju_show_units.values()` returns a dict of show_units like:

```python
juju_show_units = {
    "model_a": {"app_1": APP_1_SHOW_UNIT, "app_2": APP_2_SHOW_UNIT},
    "model_b": {"app_3": APP_3_SHOW_UNIT, "app_4": APP_4_SHOW_UNIT},
}
```

The question is whether we should call this `show_unit` or `parent`:
```python
{"app_1": APP_1_SHOW_UNIT, "app_2": APP_2_SHOW_UNIT}
```

I do see your point since `juju show-unit grafana/0 --format json` returns:
```json
{
    "loki/0": {
        "workload-version": "2.9.6",
        # snip ...
    }
}
```
and in the juju-doctor code, we run this command on all apps and aggregate them into `juju_show_units`, per model. Maybe `juju_show_units` should be renamed to `juju_models` and the for loop becomes `for show_units in juju_show_models.values():`

What do you think @wilkmar ?

wilkmar: Hi @MichaelThamm. I think you are right, 'show_units' is more adequate than 'parent' here. I updated the commit. Regarding renaming`juju_show_units` -> `juju_models`, this is only true when run online. For the offline mode (show-unit output provided as .yaml file), top level keys in this dict will be the file names for each `--show-unit` command option.

MichaelThamm (approved):
Looks good to me! Thank you for the contribution.

PR #448, opened 2025-09-12
feat: bump workload version to Grafana 12

lucabello:
## Issue
Bump the workload version from Grafana 9 to Grafana 12.

## Upgrade Notes
https://discourse.charmhub.io/t/cos-will-start-using-grafana-12-what-changed/18868


lucabello:
The Traefik issue should be unrelated to this change, as it's a charm concern (the workload shouldn't be aware of ingress at all).

Can you check if:
- removing and re-adding the `grafana:ingress traefik` relation fixes the problem
- your Traefik is working for other charms

Abuelodelanada:
> The Traefik issue should be unrelated to this change, as it's a charm concern (the workload shouldn't be aware of ingress at all).
> 
> Can you check if:
> 
>     * removing and re-adding the `grafana:ingress traefik` relation fixes the problem

I've tried that, and does not solve the issue. 


 
>     * your Traefik is working for other charms

Yes, it is working for the other charms

 
> What about normal usage of Grafana? Do you notice anything wrong?


From my manual testing, without traefik everything works fine...I did not find any issue.

Abuelodelanada:
I've modified `Grafana`'s `pebble_env` argument into a `Calllable` because when the relation with `traefik` is removed, the Pebble envs were created **before** the `TraefikRouteRequirerReadyEvent` is received. This way the `GF_SERVER_ROOT_URL` was nor updated

PR #453, opened 2025-09-26
fix: Remove `type_convert_stored` function because `cosl` owns it

MichaelThamm:
## Issue

`cosl` now owns the `type_convert_stored` function:
- https://github.com/canonical/cos-lib/pull/155

## Solution

Import this function from `cosl` to avoid code duplication.


lucabello (approved):
grafana-dashboard should be LIBPATCH 45, not 46

PR #457, opened 2025-10-03
fix:  Capture `ValueError` exception in  `_replace_uid`

Abuelodelanada:
## Issue

This PR fixes the method `_replace_uid` in `CharmedDashboard` which was reported in:

https://github.com/canonical/cos-configuration-k8s-operator/issues/128

## Solution

This method replaces the uid of the dashboard by generating one with the combination of the `charm_name` and the `real_path` of the dashboard.

When we use this method in the context of `cos-configuration-k8s` which downloads the dashboards from a git repo, the path of the dashboard is not relative to the charm_dir. Because of this, `dashboard_path.relative_to(charm_dir)` raises an `ValueError`:

```python
ValueError: '/var/lib/juju/storage/content-from-git/0/repo/grafana_dashboards/sdcore/5g_network.json' does not start with '/var/lib/juju/agents/unit-cos-configuration-0/charm'
```

The fix captures that exception and generate an uid with the `charm_name` and the `dashboard_path`

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -218,7 +218,7 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 45
+LIBPATCH = 46
 
 PYDEPS = ["cosl >= 0.0.50"]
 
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -958,10 +966,9 @@ def _replace_uid(
                 original_uid,
                 dashboard_dict["uid"],
             )
-        else:
-            logger.debug(
-                "Processed dashboard '%s': kept original uid '%s'", dashboard_path, original_uid
-            )
+        except ValueError:
```
sed-i, line 969: We should probably narrow down the ValueError to just `.relative_to()`, e.g.

```python
try:
    rel_path = ...
except ValueError:
    # Dashboard path not relative to charmdir - this is likely a dashboard from cos-configuration-k8s
    uid = DashboardPath40UID.generate(charm_name, str(dashboard_path))
else:
    uid = DashboardPath40UID.generate(charm_name, rel_path)

dashboard_dict["uid"] = uid
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -958,10 +966,9 @@ def _replace_uid(
...
+            logger.debug("dashboard_path: '%s' is no relative to charm_dir: '%s'", dashboard_path, charm_dir)
```
sed-i, line 970: When looking at debug log we probably don't care if it's relative or not?
Should we use the same debug message "replaced original uid x with y"?

PR #460, opened 2025-10-15
fix: set GF_SERVER_SERVE_FROM_SUB_PATH true if ingress available

### LINKED ISSUE #455: Dashboard redirection loop
### Bug Description

Following [revision 163](https://github.com/canonical/grafana-k8s-operator/releases/tag/rev163) where the workload version of Grafana was bumped to version 12, there appears to be an infinite dashboard redirection loop.

Trying to access the dashboard results in an infinite number of redirects where the query keeps appending onto itself in a manner such like: `/login?redirectTo=%2Flogin%3FredirectTo%3D%252Flogin%253FredirectTo%253D%25252Flogin%25253FredirectTo%25253D%2525252Flogin`

### To Reproduce

1. Deploy COS with Grafana revision 163
2. Try to access the dashboard via `curl` or a web browser.

### Environment

Juju 3.6.9
Grafana revision 163
Traefik revision 249

### Relevant log output

```shell
N/A
```

### Additional context

I'm not entirely sure if a smaller deployment can replicate this issue, but the issue does appear to have been introduced with 163, as a downgrade to 162 resolves it.

sinapah:
## Issue

This PR fixes #455. The exact cause of the issue is explained in the comments under that issue. To summarize: in the event that we are serving Grafana behind a reverse proxy (meaning `GF_SERVER_ROOT_URL` is set), we must also set `GF_SERVER_SERVE_FROM_SUB_PATH` to True. Otherwise, Grafana will not realized that it's being served behind a reverse proxy. As a result, when an unauthenticated user hits any page, Grafana will try to redirect them to the login page, but it doesn't realize that it has already hit the login page because it mistakenly takes the ingress subpath to be an internal page. Not recognizing the login page, it redirects all requests to that page to /login, creating a loop.

## Solution

This PR also performs a charm lib fetch for syncing purposes.

## Testing Instructions

Deploy:
- This Grafana charm locally.
- Traefik revision 252.
Relate the two charms.
Curl or visit any Grafana page. You should see the login page and then be successfully taken to the dashboards page. No redirect loop should happen.

### DIFF
--- a/tests/unit/test_external_url.py
+++ b/tests/unit/test_external_url.py
@@ -94,11 +94,11 @@ def test_external_url_precedence(ctx, base_state_with_model, peer_relation):
 
     # WHEN the traefik relation is removed
     with patch.object(TraefikRouteRequirer, "external_host", new=""):
-        with ctx(ctx.on.relation_departed(ingress_rel), state) as mgr:
+        with ctx(ctx.on.relation_broken(ingress_rel), state) as mgr:
             state_out = mgr.run()
             charm = mgr.charm
-            # THEN root url and subpath envs are undefined (because fqdn is a bare hostname)
-            assert get_pebble_env(charm)["GF_SERVER_SERVE_FROM_SUB_PATH"] == "True"
+            # THEN the root URL is the FQDN and Grafana does not serve from the subpath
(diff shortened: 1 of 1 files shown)


sinapah:
Changes placed on `2/edge/PR460` for testing.

Abuelodelanada (approved):
Great job sina!

PR #463, opened 2025-10-21
1/3 - Proper HA support with mysql/postgres

dstathis:
## Upcoming PRs
2/3 - Switch from traefik-route to ingress
3/3 - Service mesh support

## Issue
The current database relation supported by grafana-k8s does not have any providers and likely would not work even if it did. With this PR, we properly support mysql and postgres in a modern way.

## Solution
Add the mysql and postgres  relations and configure Grafana to use them.

### <ins>Attention</ins>
There is currently a [bug](https://github.com/grafana/grafana/issues/45639) that causes Grafana to not work with charmed mysql-k8s. The code in this PR properly supports having both relations available, but the mysql relation will cause Grafana to crash. Please advise on which path you would like me to take:
* Remove the mysql relation and any associated code completely
* Set blocked when related to mysql, thus keeping the code for managing both, so that the block can be removed when the upstream bug is resolved.
### <ins>Solution</ins>
We decided to drop mysql support.

## Testing Instructions
* Deploy grafana-k8s
* Scale up and down
* Deploy postgresql-k8s
* Relate them
* Scale up and down

### DIFF
--- a/tests/integration/test_tls_web.py
+++ b/tests/integration/test_tls_web.py
@@ -39,7 +39,8 @@ async def test_deploy(ops_test, grafana_charm):
     sh.juju.deploy(
         "self-signed-certificates", "ca", model=ops_test.model.name, channel="1/stable"
     )
-    await ops_test.model.add_relation(f"{grafana.name}:pgsql", "pgsql") # Database needed for >=2 units.
+    # Since Grafana has a scale of 2, we need a db.
+    await ops_test.model.add_relation(f"{grafana.name}:pgsql", "pgsql")
     await ops_test.model.add_relation(f"{grafana.name}:certificates", "ca")
 
     await asyncio.gather(


src/charm.py
```diff
@@ -189,12 +188,25 @@ def __init__(self, *args):
...
         )
 
+        # -- database relation
+        self._db_name = f"{self.model.name}-{self.app.name}-grafana-k8s"
+        if self.model.relations[MYSQL_RELATION] and not self.model.relations[PGSQL_RELATION]:
+            self._db = DatabaseRequires(self, relation_name=MYSQL_RELATION, database_name=self._db_name)
+            self._db_type = "mysql"
+        elif self.model.relations[PGSQL_RELATION]:
+            self._db = DatabaseRequires(self, relation_name=PGSQL_RELATION, database_name=self._db_name)
+            self._db_type = "postgres"
```
sed-i: Can we be certain that the conditional construction of `self._db` isn't racy? For example
- The "joined" of pg happens before the "broken" of mysql -- in this case `self._db` will be pg, and we'd miss relation hooks for mysql.
- Is switching between litestream to the other databases lossy? I.e. would the admin need to manually recover the admin password?

Would it make sense to always instantiate e.g. `self.db = {"litestream": ..., "postgres": ..., ...}`?

src/charm.py
```diff
@@ -189,12 +188,25 @@ def __init__(self, *args):
             refresh_event=self.on.grafana_pebble_ready,  # pyright: ignore
         )
 
+        # -- database relation
+        self._db_name = f"{self.model.name}-{self.app.name}-grafana-k8s"
```
sed-i: Technically this db isn't unique, because we cannot guarantee the db is always dedicated / in-model.
- Should we use a method from `self._topology`?
- Would the db lib give us a default db name so we won't need to specify a db name at all?

src/charm.py
```diff
@@ -418,8 +431,7 @@ def _provision_own_dashboard(self) -> bool:
     @property
     def _enable_external_db(self) -> bool:
         """Only consider a DB connection if we have config info."""
-        rel = self.model.get_relation(DATABASE_RELATION)
-        return len(rel.units) > 0 if rel is not None else False
+        return bool(self.model.get_relation(MYSQL_RELATION) or self.model.get_relation(PGSQL_RELATION))
```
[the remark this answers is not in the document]
dstathis: Done.

src/charm.py
sed-i: Model uuid would make it unique. Any reason not to use it? I.e. does the admin care how the db is named?

src/charm.py
```diff
@@ -418,8 +423,7 @@ def _provision_own_dashboard(self) -> bool:
     @property
     def _enable_external_db(self) -> bool:
         """Only consider a DB connection if we have config info."""
-        rel = self.model.get_relation(DATABASE_RELATION)
-        return len(rel.units) > 0 if rel is not None else False
+        return bool(self.model.get_relation(PGSQL_RELATION))
```
sed-i: This seems redundant if we now require a db relation and do not have litestream.
I.e. iiuc the charm would be in blocked (broken) state without a db relation anyway.

src/charm.py
```diff
@@ -479,11 +483,19 @@ def _reconcile(self):
...
         self._reconcile_relations()
         self._grafana_service.reconcile()
-        self._litestream.reconcile()
         self._reconcile_tls_config()
 
+    def _check_wrong_relations(self) -> Optional[StatusBase]:
+        """Check that relations are configured properly."""
+        relations = self.model.relations
+        if not relations[PGSQL_RELATION] and self.app.planned_units() > 1:
+            return BlockedStatus("Scale > 1 requires pgsql relation.")
```
sed-i: Should we bother managing two different databases for the single unit case?
Imagine an admin set up a bunch of things via api, plugins etc, now they scale up, do we want to maintain logic for crossing over between one db to another?
But maybe I'm missing something.

dstathis: The idea here is that cos-lite should still work without Postgres as it's a bit heavy. Data is not migrated when a relation is made, however, the data persists in the SQLite database if you remove the relation again. (And same for the other way around) So you will never lose your data as you can simply go back to the previous database.

src/charm.py
dstathis: At scale 1, we still allow using SQLite without a db relation.

src/charm.py
sed-i: Oh great, this is fine then.

src/charm.py
```diff
@@ -189,12 +186,22 @@ def __init__(self, *args):
...
         )
 
+        # -- database relation
+        self._db_name = f"{self._topology.application}-grafana-k8s-{self._topology.model_uuid}"
+        if self.model.relations[PGSQL_RELATION]:
+            self._db = DatabaseRequires(self, relation_name=PGSQL_RELATION, database_name=self._db_name)
+            self._db_type = "postgres"
+        else:
+            self._db = None
+            self._db_type = "sqlite3"
```
Abuelodelanada: Minor change: remove unnecessary `else`

```suggestion
        self._db = None
        self._db_type = "sqlite3"

        if self.model.relations[PGSQL_RELATION]:
            self._db = DatabaseRequires(self, relation_name=PGSQL_RELATION, database_name=self._db_name)
            self._db_type = "postgres"
```

dstathis:
> # Manual test
> ## Deploy both charms, but did not relate it.
> ```shell
> Model  Controller  Cloud/Region        Version  SLA          Timestamp
> cos    microk8s    microk8s/localhost  3.6.11   unsupported  17:23:33-03:00
> 
> App         Version  Status  Scale  Charm           Channel  Rev  Address         Exposed  Message
> grafana     12.0.2   active      1  grafana-k8s                0  10.152.183.132  no       
> postgresql  16.10    active      1  postgresql-k8s  16/edge  687  10.152.183.25   no       
> 
> Unit           Workload  Agent  Address       Ports  Message
> grafana/0*     active    idle   10.1.102.205         
> postgresql/0*  active    idle   10.1.102.238         Primary
> 
> Integration provider       Requirer                   Interface         Type  Message
> grafana:grafana            grafana:grafana            grafana_peers     peer  
> grafana:replicas           grafana:replicas           grafana_replicas  peer  
> postgresql:database-peers  postgresql:database-peers  postgresql_peers  peer  
> postgresql:restart         postgresql:restart         rolling_op        peer  
> postgresql:upgrade         postgresql:upgrade         upgrade           peer  
> ```
> 
> Grafana works OK:
> 
> image
> ## Scale Grafana
> Grafana goes to `blocked` status
> 
> ```shell
> Model  Controller  Cloud/Region        Version  SLA          Timestamp
> cos    microk8s    microk8s/localhost  3.6.11   unsupported  17:27:03-03:00
> 
> App         Version  Status   Scale  Charm           Channel  Rev  Address         Exposed  Message
> grafana     12.0.2   blocked      2  grafana-k8s                0  10.152.183.132  no       missing database relation
> postgresql  16.10    active       1  postgresql-k8s  16/edge  687  10.152.183.25   no       
> 
> Unit           Workload  Agent  Address       Ports  Message
> grafana/0*     blocked   idle   10.1.102.205         missing database relation
> grafana/1      blocked   idle   10.1.102.213         missing database relation
> postgresql/0*  active    idle   10.1.102.238         Primary
> 
> Integration provider       Requirer                   Interface         Type  Message
> grafana:grafana            grafana:grafana            grafana_peers     peer  
> grafana:replicas           grafana:replicas           grafana_replicas  peer  
> postgresql:database-peers  postgresql:database-peers  postgresql_peers  peer  
> postgresql:restart         postgresql:restart         rolling_op        peer  
> postgresql:upgrade         postgresql:upgrade         upgrade           peer  
> ```
> 
> ...but Grafana (`grafana/0`) still works:
> 
> image
> On the other hand, `grafana/1` works too, but I cannot log in with the password returned by `get-admin-password` action.
> 
> image
> Both units return the same password:
> 
> ```shell
> ╭─ubuntu@charm-dev-362 ~/repos/grafana-k8s-operator ‹remove_lightstream●› [microk8s:cos]
> ╰─$ juju run grafana/1 get-admin-password -m cos
> Running operation 3 with 1 task
>   - task 4 on unit-grafana-1
> 
> Waiting for task 4...
> admin-password: bs3AMKVrKMaS
> url: http://grafana-1.grafana-endpoints.cos.svc.cluster.local:3000
> 
> ╭─ubuntu@charm-dev-362 ~/repos/grafana-k8s-operator ‹remove_lightstream●› [microk8s:cos]
> ╰─$ juju run grafana/0 get-admin-password -m cos
> Running operation 5 with 1 task
>   - task 6 on unit-grafana-0
> 
> Waiting for task 6...
> admin-password: bs3AMKVrKMaS
> url: http://grafana-0.grafana-endpoints.cos.svc.cluster.local:3000
> ```
> 
> ## Relate Grafana to PostgreSQL
> When I relate both,
> 
> ```shell
> juju relate grafana postgresql
> ```
> 
> the only valid relation between them is `grafana-dashboard`:
> 
> ```shell
> Model  Controller  Cloud/Region        Version  SLA          Timestamp
> cos    microk8s    microk8s/localhost  3.6.11   unsupported  17:40:24-03:00
> 
> App         Version  Status   Scale  Charm           Channel  Rev  Address         Exposed  Message
> grafana     12.0.2   blocked      2  grafana-k8s                0  10.152.183.132  no       missing database relation
> postgresql  16.10    active       1  postgresql-k8s  16/edge  687  10.152.183.25   no       
> 
> Unit           Workload  Agent  Address       Ports  Message
> grafana/0*     blocked   idle   10.1.102.205         missing database relation
> grafana/1      blocked   idle   10.1.102.213         missing database relation
> postgresql/0*  active    idle   10.1.102.238         Primary
> 
> Integration provider          Requirer                   Interface          Type     Message
> grafana:grafana               grafana:grafana            grafana_peers      peer     
> grafana:replicas              grafana:replicas           grafana_replicas   peer     
> postgresql:database-peers     postgresql:database-peers  postgresql_peers   peer     
> postgresql:grafana-dashboard  grafana:grafana-dashboard  grafana_dashboard  regular  
> postgresql:restart            postgresql:restart         rolling_op         peer     
> postgresql:upgrade            postgresql:upgrade         upgrade            peer 
> ```

Hm that is very strange. Can you check that the charm you are building contains this in charmcraft.yaml?
20251024_23h22m09s_grim

tests/unit/test_external_url.py
```diff
@@ -8,7 +8,8 @@
 from typing import Dict
 from unittest.mock import patch
 from pytest import fixture, mark
-from charms.traefik_k8s.v0.traefik_route import TraefikRouteRequirer
+#from charms.traefik_k8s.v0.traefik_route import TraefikRouteRequirer
```
MichaelThamm: Leftover?

charmcraft.yaml
```diff
@@ -48,18 +43,17 @@ requires:
   grafana-auth:
     interface: grafana_auth
     optional: true
-  database:
-    interface: db
+  pgsql:  # Do not call this endpoint "database" because we may want to add mysql support at some point.
+    interface: postgresql_client
     limit: 1
     optional: true
```
MichaelThamm:
```suggestion
    optional: true
    description: This endpoint is specific to a PostgreSQL database backend which provides high availability for data storage. Using this endpoint overrides Grafana's usage of the embedded sqlite3 database for data storage.
```

You might have to format the string value, but you get the idea. Please double check if this statement is correct. I got it from:
- https://grafana.com/docs/grafana/latest/setup-grafana/set-up-for-high-availability/

tests/integration/test_tls_web.py
```diff
@@ -34,14 +33,18 @@ async def test_deploy(ops_test, grafana_charm):
         resource=[f"{k}={v}" for k, v in grafana_resources.items()],
         num_units=2,
     )
+    sh.juju.deploy(
+        "postgresql-k8s", "pgsql", model=ops_test.model.name, channel="14/stable", trust=True
+    )
     sh.juju.deploy(
         "self-signed-certificates", "ca", model=ops_test.model.name, channel="1/stable"
     )
+    await ops_test.model.add_relation(f"{grafana.name}:pgsql", "pgsql") # Database needed for >=2 units.
```
MichaelThamm:
```suggestion
    # Since Grafana has a scale of 2, we need a db.
    await ops_test.model.add_relation(f"{grafana.name}:pgsql", "pgsql")
```

MichaelThamm (approved):
This looks great Dylan!

Please make sure you sync with Jose about [this issue](https://github.com/canonical/grafana-k8s-operator/pull/463#issuecomment-3444786653) and merge [this PR](https://github.com/canonical/observability-stack/pull/149) after you merge this one. Assuming these are resolved, you have my approval to merge.

PR #467, opened 2025-10-22
fix: remove tls router and middlewares

### LINKED ISSUE #468: Dashboard redirection loop when not using TLS due to prefix being stripped by traefik route middleware
### Bug Description

This issue is a follow up of #455, reported by @gaborbk.

This issue is in conjunction with (#469).

We are currently serving Grafana from the ingress subpath when ingress is available through Traefik.
https://github.com/canonical/grafana-k8s-operator/blob/27e185858abcccf0134a79650930cf19a0b7f76a/src/grafana.py#L131-L133

However, we are also using routers and middlewares with Traefik. Specifically, the non-TLS router has the following middleware:
https://github.com/canonical/grafana-k8s-operator/blob/27e185858abcccf0134a79650930cf19a0b7f76a/src/charm.py#L289-L302

Hence, when requests arrive at http without TLS, the subpath is removed. Since we are serving Grafana from the subpath when using ingress, this leads to a redirection loop, where the request never even gets to Grafana because it's being redirected to the subpath, resulting in a loop.

### To Reproduce

1. Deploy Grafana revision 169
2. Deploy Traefik revision 254.
3. Relate them: `juju relate traefik grafana:ingress`.
4. `curl -vkL http://<traefik-address>/<model-name>-<grafana-name>`
5. The step above should result in a redirection loop.
6. curl -vkL https://<traefik-address>/<model-name>-<grafana-name>
7. You'll notice that the command in (6) will get a 200. The reason is because due to a separate issue (explained in #469), the TLS router uses no middlewares. Since Grafana is serving from subpath, it recognizes the "unstripped" subpath and things work, even though we are not truly using TLS.

### Environment

Running inside Multipass VM. Using Juju 3.6.10 and Ck8s 1.33-classic/stable. 

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

### LINKED ISSUE #469: Grafana uses TraefikRouteProvider with raw=False
### Bug Description

This issue partially contributes to #468. It is also related to [#548 in Traefik](https://github.com/canonical/traefik-k8s-operator/issues/548) and [508 in Traefik](https://github.com/canonical/traefik-k8s-operator/issues/508).

When using traefik route, the `TraefikRouteRequirer` has a `raw` parameter, a boolean which defaults to false when `TraefikRouteRequirer` is instantiated. When this is false, Traefik expects there to only be one non-TLS router in relation data. It then makes a call to `process_routes`, which essentially duplicates that non-TLS router by appending `-tls` to it and stripping its middlewares.

https://github.com/canonical/traefik-k8s-operator/blob/1ec7fcb002149dfdf154a4a4570f8a42db4433d8/src/charm.py#L1219-L1223

Hence why the TLS middleware in Grafana seems to be missing its middlewares and why the TLS router is duplicated (because since raw=False, Traefik thinks it has to create a TLS router for the TLS router that Grafana has written to relation data as well).
This is because we write both a non-TLS and a TLS router.
https://github.com/canonical/grafana-k8s-operator/blob/27e185858abcccf0134a79650930cf19a0b7f76a/src/charm.py#L296-L317
```
http:
  middlewares:
    juju-sidecar-noprefix-observable-model-graf:
      stripPrefix:
        forceSlash: false
        prefixes:
        - /observable-model-graf
  routers:
    juju-observable-model-graf-router:
      entryPoints:
      - web
      middlewares:
      - juju-sidecar-noprefix-observable-model-graf
      rule: PathPrefix(`/observable-model-graf`)
      service: juju-observable-model-graf-service
    juju-observable-model-graf-router-tls:
      entryPoints:
      - websecure
      rule: PathPrefix(`/observable-model-graf`)
      service: juju-observable-model-graf-service
      tls: {}
    juju-observable-model-graf-router-tls-tls:
      entryPoints:
      - websecure
      rule: PathPrefix(`/observable-model-graf`)
      service: juju-observable-model-graf-service
      tls: {}

```

To fix this, we have two options:
### 1. Set raw=True
This can be useful if we want to control the middlewares. Since Traefik strips the middlewares, this option is significant if we want to preserve our middlewares.

### 2. Only write a non-TLS router, without using middlewares (better)
We do not actually need to use middlewares in Grafana. If we correctly use `GF_serve_from_subpath`, we do not need to have a middleware which removes the ingressed subpath. In fact, if strip the prefix and incorrectly set the subpath serve env to True, this results in a redirection loop, where a request never makes its way to Grafana.
I recommend that we remove our middlewares and only write one router. Traefik takes care of creating the TLS router then.

### To Reproduce

See attached issues.

### Environment

N/A

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

sinapah:
## Issue

Fixes #468 and fixes #469.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -61,7 +61,7 @@ def patch_pylibjuju_series_2404():
 
     yield
 
-    juju.utils.ALL_SERIES_VERSIONS["noble"]
+    del juju.utils.ALL_SERIES_VERSIONS["noble"]
     del juju.utils.UBUNTU_SERIES["noble"]
 
 


src/charm.py
```diff
@@ -272,47 +272,12 @@ def _ingress_config(self) -> dict:
...
                 "entryPoints": ["web"],
                 "rule": f"PathPrefix(`/{external_path}`)",
-                "middlewares": list(middlewares.keys()),
-                "service": "juju-{}-{}-service".format(self.model.name, self.app.name),
-            },
-            "juju-{}-{}-router-tls".format(self.model.name, self.model.app.name): {
-                "entryPoints": ["websecure"],
-                "rule": f"PathPrefix(`/{external_path}`)",
-                "middlewares": list(middlewares.keys()),
+                "middlewares": [],
```
sed-i: Should we add a comment here about not needing middlewares here, leaving it to traefik?

sed-i (commented):
This is great!
Just need to correctly capture "internal TLS".

tests/integration/test_ingress.py
```diff
@@ -0,0 +1,181 @@
...
+
+@pytest.mark.abort_on_fail
+async def test_no_tls(ops_test):
+    """Test whether we are able to reach Grafana via Traefik when using no TLS.
+
+    This means that both Traefik and Grafana are only using HTTP.
+    """
+    traefik_address = await get_traefik_url(ops_test, traefik_app_name=traefik_app_name)
+
+    assert "http://" in traefik_address
```
sed-i: Nit:

```suggestion
    assert traefik_address.startswith("http://")
```

tests/integration/test_ingress.py
```diff
@@ -0,0 +1,181 @@
...
+    grafana_address = grafana_address.replace("https://", "http://")
+    fetch_with_retry(url=grafana_address, expected_status=301, follow_redirects=False)
+
+    # And if we allow redirects, we should get a 200
+    fetch_with_retry(url=grafana_address, expected_status=200)
+
+@pytest.mark.abort_on_fail
+async def test_external_tls(ops_test):
+    # Remove relation between Grafana and SSC. This means that Grafana no loger uses TLS.
+    # TLS is terminated now at Traefik and traffik between Traefik and Grafana is unencrpyted.
```
sed-i:
```suggestion
    # TLS is terminated now at Traefik, so traffic between Traefik and Grafana is unencrpyted.
```

Issue #342, opened 2024-07-05
Duplicate dashboards remove write access to database


cbartz:
### Bug Description

We have a deployment where a `grafana-agent` and a `grafana-agent-k8s` application are integrated with grafana:grafana-dashboard . Both provide dashboards with the title "System Resources" (https://github.com/canonical/grafana-agent-operator/blob/6eb1938895f5c31704917832d7d84ca9afaed799/src/grafana_dashboards/node-exporter-full.json#L23196 and https://github.com/canonical/grafana-agent-operator/blob/d77e500625a75318f16eff5fad7bb8878da141fc/src/grafana_dashboards/node-exporter-full.json#L23196). This leads to warnings like 
```
2024-07-04T09:05:19.197Z [grafana] logger=provisioning.dashboard t=2024-07-04T09:05:19.197743344Z level=warn msg="dashboard title is not unique in folder" orgId=1 title="System Resources" folderID=0 times=2 providers=[Default]
2024-07-04T09:05:19.197Z [grafana] logger=provisioning.dashboard t=2024-07-04T09:05:19.197764885Z level=warn msg="dashboards provisioning provider has no database write permissions because of duplicates" provider=Default orgId=1
2024-07-04T09:05:19.223Z [grafana] logger=provisioning.dashboard type=file name=Default t=2024-07-04T09:05:19.223731211Z level=warn msg="Not saving new dashboard due to restricted database access" provisioner=Default file=/etc/grafana/provisioning/dashboards/juju_grafana-agent-k8s_0def0c2.json folderId=0
```

and the provisioning of some other dashboards failed in our production environment (they did not appear in Grafana). Grafana seems to remove write access to the database in this case: https://github.com/grafana/grafana/issues/43530

Besides this, we also have an application deployed with three units, which provides two times the same dashboard in the relational data, which leads to

```
2024-07-04T09:05:19.197Z [grafana] logger=provisioning.dashboard t=2024-07-04T09:05:19.197753914Z level=warn msg="dashboard title is not unique in folder" orgId=1 title="Synapse Operator" folderID=0 times=2 providers=[Default]
```

A `juju show unit grafana/0` gives:

```
 "37": [{"id": "grafana-dashboard:37/file:synapse", "original_id": "file:synapse"

        "juju_topology": {"model": "prod-synapse-k8s", "model_uuid": "cf54a174-d8f9-4a6a-8e43-2a17d555c60b",
        "application": "synapse", "unit": "synapse/0"}
[...]

PR #484, opened 2025-11-22
OBC-1398 Provision only the latest version per dashboard UID

sed-i:
## Issue
When charmed dashboards are modified, the convention is to bump the `.version` field. That's also what the grafana UI does when you make changes to a dashboard and save it.

If two revisions of a charm are related to grafana, the dashboards they send over reldata may differ.
Currently, both will be provisioned (written to disk), but in the grafana UI only one appears, and it's unclear which of the dashboards it is.

Moreover, in the UI, grafana [resets](https://grafana.com/docs/grafana/next/administration/provisioning/#make-changes-to-a-provisioned-dashboard) the `.version` of provisioned dashboards to `1`.

## Solution
Write to disk the dashboards with the highest version per uid.
This also means that charm devs should be responsible for bumping the version field of the dashboards they modify. 

Since Grafana 12 will reset the `.version` to `1` on export, and since charm devs are not likely to remember to manually bump it back to N+1 after they export the dashboard with new changes, this means that taking the latest version may end up giving you an older copy of the dashboard.

There isn't much else we can do about deduplication. We can't reliably enrich the json with the charm revision. The only sensible thing seems to be to emit a warning when multiple versions exist for the same uid.

All in all I think this is a positive change because now the behavior is well defined - keep highest version per uid - whereas before this change you can't easily tell in advance which version will be picked up by grafana and displayed in the UI. 

## Context
This is a follow-up on:
- https://github.com/canonical/observability/pull/206
- https://github.com/canonical/grafana-k8s-operator/issues/342
- https://github.com/canonical/grafana-k8s-operator/pull/363

## Testing Instructions
- Manually pack avalanche, once as-is, and once with bumping the `version` of the avalanche dashboard.
- Deploy both charms and relate to grafana.
- Confirm with `juju ssh` which dashboard is written to disk. Should be with the higher version.
- Play with removing/re-adding the dashboards relation to see how the dashboard files alternate on disk.

```mermaid
graph LR
graf ---|metrics-endpoint| prom
graf ---|grafana-source| prom
av ---|metrics-endpoint| prom
av ---|grafana-dashboard| graf
av10 ---|grafana-dashboard| graf
```

```yaml
bundle: kubernetes
applications:
  av:
    charm: local:avalanche-k8s-10
    scale: 1
    constraints: arch=amd64
  av10:
    charm: local:avalanche-k8s-11
    scale: 1
    constraints: arch=amd64
  graf:
    charm: local:grafana-k8s-1
    resources:
      grafana-image: 77
      litestream-image: 46
    scale: 1
    options:
      reporting_enabled: false
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  prom:
    charm: prometheus-k8s
    channel: 2/stable
    revision: 272
    resources:
      prometheus-image: 152
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
relations:
- - prom:metrics-endpoint
  - graf:metrics-endpoint
- - graf:grafana-source
  - prom:grafana-source
- - av:metrics-endpoint
  - prom:metrics-endpoint
- - av:grafana-dashboard
  - graf:grafana-dashboard
- - av10:grafana-dashboard
  - graf:grafana-dashboard
```

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -1679,12 +1679,12 @@ class GrafanaDashboardConsumer(Object):
                 key = obj.get("dashboard_uid")
                 if key is None or str(key).strip() == "":
                     # At this point, we assume that a `.uid` is present so we do not render a fallback identifier here. Instead, we omit it.
-                    logger.error("Dashboard '%s' from relation id '%s' is missing a '.uid' field; omitted", obj["dashboard_title"], obj["relation_id"])
+                    logger.error("dashboard '%s' from relation id '%s' is missing a '.uid' field; omitted", obj["dashboard_title"], obj["relation_id"])
                     continue
 
                 if key in d:
                     d[key] = max(d[key], obj, key=lambda o: (o["dashboard_version"], o["relation_id"], o["content"]))
-                    logger.warning("Deduplicate dashboard '%s' (uid '%s') - kept version '%s' from relation id '%s'", d[key]["dashboard_title"], d[key]["dashboard_uid"], d[key]["dashboard_version"], d[key]["relation_id"])
+                    logger.warning("deduplicate dashboard '%s' (uid '%s') - kept version '%s' from relation id '%s'", d[key]["dashboard_title"], d[key]["dashboard_uid"], d[key]["dashboard_version"], d[key]["relation_id"])
                 else:
                     d[key] = obj
 


src/grafana.py
```diff
@@ -281,7 +283,28 @@ def _reconcile_dashboards(self):
...
+                uid = dash_dict["uid"]
+                title = dash_dict.get("title")
+
+                if not (ver := dash_dict.get("version")):
+                    ver = 0
+                    logger.warning("Dashboard '%s' (uid '%s') is missing a '.version' field; using '0' as fallback", title, uid)
+                ver = int(ver)
+
+                if uid in latest_versions and ver > latest_versions[uid].version:
+                    old_title = json.loads(latest_versions[uid].dashboard["content"]).get("title")
```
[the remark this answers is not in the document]
sed-i: @copilot add "title" to the namedtuple (not the entire json parsed dashboard). This way we won't need `old_title`.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1636,29 +1635,60 @@ def _remove_all_dashboards_for_relation(self, relation: Relation) -> None:
...
+        as_dict = json.loads(decompressed)
+
+        dashboard_title = as_dict.get("title", "")
+        dashboard_uid = as_dict.get("uid", "")
+
+        try:
+            dashboard_version = int(as_dict["version"])
+        except (KeyError, ValueError):
+            logger.warning("Dashboard '%s' (uid '%s') is missing a '.version' field or is invalid (must be integer); using '0' as fallback", dashboard_title, dashboard_uid)
+            dashboard_version = 0
```
Abuelodelanada: Wouldn't be valuable to know exactly which error are we facing?

```suggestion
        dashboard_version = 0
        try:
            dashboard_version = int(as_dict["version"])
        except KeyError:
            logger.warning("dashboard '%s' (uid '%s') is missing a '.version' field", dashboard_title, dashboard_uid)
        except ValueError:
            logger.warning("dashboard '%s' (uid '%s') '.version' field is invalid (must be integer); using '0' as fallback", dashboard_title, dashboard_uid)
```

Abuelodelanada (approved):

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: I started like that, and then realized that the action taken wouldn't change between the two cases: need to inspect and fix the version field.

PR #499, opened 2026-01-21
feat(config): add charm config option for custom grafana config

### LINKED ISSUE #487: Surface SMTP config options as juju config
### Enhancement Proposal

We need to surface a way for users to configure the SMTP options of Grafana using Juju config.

sed-i:
## Issue
Users are unable to use the "send email" feature of Grafana, because the charm doesn't surface that config.

## Solution
- [x] Add a general purpose "custom config" option to the charm, currently accepting only the [`[smtp]` section](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#smtp). This approach is similar to how we handle arbitrary config in alertmanager, script exporter.
- [x] Add a `.get_status()` and use it in collect-unit-status to indicate invalid config.
- [ ] Auto update the settings with a default "contact point" via [`PUT /api/admin/settings`](https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/admin/#update-settings).

Fixes #487.

## Context
After the smtp config is in, there is a manual step needed for [adding a "contact point"](https://grafana.com/docs/grafana/latest/alerting/configure-notifications/manage-contact-points/integrations/configure-email/). Afaict, it should be possible to update it programmatically with [`PUT /api/admin/settings`](https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/admin/#update-settings), but that doc page hints that the endpoint is only available on grafana enterprise. Need to confirm.

### DIFF
--- a/src/custom_ini_config.py
+++ b/src/custom_ini_config.py
@@ -33,7 +33,7 @@ class _SMTPSection(BaseModel):
     key_file: Optional[str] = None
     """File path to a key file, default is empty."""
 
-    skip_verify: bool = False
+    skip_verify: Optional[bool] = None
     """Verify SSL for SMTP server, default is false."""
 
     from_address: Optional[EmailStr] = None


src/custom_ini_config.py
```diff
@@ -0,0 +1,84 @@
...
+    "Name to be used as client identity for EHLO in SMTP conversation, default is <instance_name>."""
+
+    startTLS_policy: Optional[Literal["OpportunisticStartTLS", "MandatoryStartTLS", "NoStartTLS"]] = None  # noqa: N815
+    """Either OpportunisticStartTLS, MandatoryStartTLS, NoStartTLS, or empty. Default is empty."""
+
+    enable_tracing: Optional[bool] = None
+    "Enable trace propagation in email headers, using the traceparent, tracestate and (optionally) baggage fields. Default is false. To enable, you must first configure tracing in one of the tracing.opentelemetry.* sections."""
+
+
+def validate(ini_sections: Optional[str]):
```
sed-i: @copilot add unit tests for this function.
Unit tests go under tests/unit.
All the new unit tests should go into a single file, `test_custom_ini_config.py`.

Implement the following tests:
```python
# WHEN validate called without args
# THEN returns None

# WHEN validate called with "hello"
# THEN ValueError is raised

# WHEN validate called with "[section]\nkey = value"
# THEN ValueError is raised

# WHEN validate called with "[smtp]\nenabled = true"
# THEN returns None

# WHEN validate called with "[smtp]\nenabled = true\n\invalid = value"
# THEN ValueError is raised
```

PR #502, opened 2026-01-29
add role_attribute_path

sed-i:
## Issue
Via @weiiwang01.

Admins want to enable oauth in the Grafana charm and assign the role "editor" to every user who logs in via OAuth, instead of the default "viewer".

## Solution
Add `admin_roles`, `editor_roles` config options that maps into the grafana.ini config.

TODO:
- [x] When refreshing a charm with this change to 2/stable, make sure the grafana.ini is updated and does not contain the roles entries.

References:
- https://grafana.com/docs/grafana/latest/setup-grafana/configure-access/configure-authentication/generic-oauth/#configuration-options
- https://jmespath.org/examples.html

## Context
[Matrix thread](https://matrix.to/#/!nHXpRkcSNJHlHdUGbQ:ubuntu.com/$YnDLTLgcM3mQODYJHVGJ6tVoBIERqFkiaagRoiNVHhQ?via=ubuntu.com&via=matrix.org&via=mozilla.org).

## Testing Instructions
See @nsklikas message [below](https://github.com/canonical/grafana-k8s-operator/pull/502#issuecomment-4097226063).

### DIFF
--- a/tests/integration/test_grafana_oauth.py
+++ b/tests/integration/test_grafana_oauth.py
@@ -42,15 +42,15 @@ grafana_resources = {
 }
 
 
-pytest.skip(
+skip_reason = (
     "Grafana goes into error state because the ouath_tools lib is outdated and results in schema validation errors for cert_transfer: "
     "'Make sure not to interact with the databags except using the public methods in the provider library and use version V1.'"
     "For future ref, in jimm they use a literal bundle instead of oauth_tools: "
-    "https://github.com/canonical/jimm-k8s-operator/blob/v3/tests/integration/identity-bundle.yaml",
-    allow_module_level=True,
+    "https://github.com/canonical/jimm-k8s-operator/blob/v3/tests/integration/identity-bundle.yaml"
 )
 
 
+@pytest.mark.skip(reason=skip_reason)
 @pytest.mark.skip_if_deployed
 @pytest.mark.abort_on_fail
 async def test_build_and_deploy(
@@ -101,6 +101,7 @@ async def test_build_and_deploy(
     )
 
 
+@pytest.mark.skip(reason=skip_reason)
 async def test_oauth_login_with_identity_bundle(
     ops_test: OpsTest,
     page: Page,


shipperizer (approved):

charmcraft.yaml
```diff
@@ -257,3 +257,19 @@ config:
         Ref: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#analytics
       type: boolean
       default: true
+    role_attribute_path:
```
nsklikas: thought: This requires the user to:
- Know what the attribute path is
- Provide a complex string via juju config (and juju configs are very restricting)

IMO it would make most sense to have 3 configurations:
- `admin_roles`
- `editor_roles`
- `viewer_roles`

Grafana would then get the `group` attributes path from the hydra databag and it would use these config to construct the JMESPath. eg:
```
hydra_path = "groups"
path = f"contains({hydra_path}[*], '{admin_roles}') && 'Admin' || contains({hydra_path}[*], '{editor_roles}') && 'Editor' || contains({hydra_path}[*], '{viewer_roles}') && 'Viewer'"
```

sed-i: I agree, and this sounds like a good idea @nsklikas, but I'm not familiar with JMESPath at all so not sure.
Would you be able to elaborate? Are the proposed `*_roles` config options strings?
Would you like branch off of this one and submit a merge request into this branch?

nsklikas: @sed-i I pushed an untested commit with a barebone implementation at https://github.com/nsklikas/grafana-k8s-operator/commit/17f3469f9d917361e817b5e03a8223b71b377c7a. Needs some work, but it should make it clear what I am suggesting. There are some annoying edge cases about what characters should allow in the juju config (e.g. should we allow any of these characters: ` ,"'\[]()`?)

NOTE: hydra does not provide the name of the `group` claim in the databag right now, perhaps we should add it (if groups are enabled)

src/grafana_config.py
```diff
@@ -55,6 +66,23 @@ def auth_env_config(self) -> Any:
...
+    @property
+    def role_attribute_path(self) -> Optional[str]:
+        """Generate role attribute path."""
+        group_claim_path  = "groups[*]"
+        if not self._admin_roles and not self._editor_roles:
+            return None
+
+        role_paths = []
+        for admin_role in self._admin_roles:
+            role_paths.append(f"contains({group_claim_path}, {admin_role}) && 'Admin'")
```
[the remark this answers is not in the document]
shipperizer: need to add single quotes and at the same time polish input to avoid that

PR #540, opened 2026-04-29
feat: migrate charm-tracing to ops[tracing]

lucabello:
## Summary

Migrate charm tracing from the old `TracingEndpointRequirer` to the new `ops_tracing.Tracing` approach (installed via `ops[tracing]`). Workload tracing continues to use `TracingEndpointRequirer`.

### Changes
- Added `ops_tracing.Tracing` instantiation for charm tracing
- Removed charm-tracing `TracingEndpointRequirer` and manual `set_destination` call
- Consolidated dependencies to use `ops[tracing]>=2.17`
- Added `cargo` to build-packages
- Updated uv.lock

### Motivation
The `ops_tracing.Tracing` object (from `ops[tracing]`) is the recommended way to add charm tracing. It handles destination configuration automatically via the relation, removing the need for manual `set_destination` calls.


src/charm.py
```diff
@@ -154,8 +154,10 @@ def __init__(self, *args):
                 self.on.update_status,
             ],
         )
-        self.charm_tracing = TracingEndpointRequirer(
-            self, relation_name="charm-tracing", protocols=["otlp_http"]
+        self.tracing = ops_tracing.Tracing(
```
mmkay: I'd call it `self.charm_tracing` to match the convention for the workload.

mmkay (approved):

src/charm.py
lucabello: Good point!

Abuelodelanada (approved):
LGTM

PR #543, opened 2026-05-06
feat!: long DB name

### LINKED ISSUE #542: Grafana charm sends an overly long database name to PostgreSQL / PgBouncer, exceeding their 63-character limit
### Bug Description

When Grafana relates to PostgreSQL or PgBouncer, it sends the desired database name without taking into account the hard limit of 63 total characters enforced by Postgres. As PgBouncer adds a `_readonly` suffix for read replicas, the database name Grafana sends in its relation must be below 63 - 9 = 54 characters, including the `grafana` prefix as well - please see https://github.com/canonical/pgbouncer-k8s-operator/issues/750 for details.

### To Reproduce

1. Deploy Grafana (3 units), PostgreSQL and Pgbouncer
2. Relate them
3. When the units restart due to infrastructure instability, eventually Pgbouncer's Pebble services go into error complaining about the database name being too long. This can be seen in the generated INI file as well, as shown in https://github.com/canonical/pgbouncer-k8s-operator/issues/750

### Environment

Juju 3.6.14, Canonical K8s, latest Grafana and PgBouncer chamrs.

### Relevant log output

```shell
Please see https://github.com/canonical/pgbouncer-k8s-operator/issues/750
```

### Additional context

_No response_

sinapah:
## Issue

Fixes #542.

## Solution

It shortens the DB name by removing the static `grafana-k8s` string. Now, we always add the the model UUID first and we append the name of the charm. If the combo of the UUID and name exceeds 54 chars, we cap the string at 54 chars.

## Context

I've started a convo here with Data Platform to ask about the options of users for migrating from the old DB to the new one. There is a bit of a concern about what happens to users who refresh their Grafana charms that are already related to PG to a revision containing this change. The removal of `grafana-k8s` from the DB name will result in the creation of a new DB without that string after refresh. All plugins as well as alerts and dashboards created by the user through the Grafana UI and stored in the DB will remain in the old DB, but Grafana will be using the new one. The same is true for user auth info. With regards to alerts and dashboards, the risk is **probably** minimal as most users rely on COS Config to supply custom assets. WRT to login info, the disturbance will be greater.
In envs where the user has operator level access to the PG primary, they can perform the migration using `psql`, but that's not a guarantee necessarily in environments relying on DBaaS.

## Upgrade Notes

Once this PR goes in, users who previously had the charm deployed and related to PGbouncer (and by extension Postgresql) will need to alter their schema so that the data in the DB with the old name "migrates" to the correct, new name expected by Grafana.


lucabello:
Is this a `feat!:` as it's a breaking change for users maybe?

MichaelThamm:
If we are shortening the DB name already, then I would recommend also shortening the model UUID to be a short UUID, like we do in alert rules. This will further guard us from long app names.

tests/integration/test_postgresql.py
sed-i, line 1: Question (non-blocking): see if we should merge this test with [`test_multiple_units`](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/integration/test_multiple_units.py)?

sed-i (approved):
Great!

sed-i:
Also: please update the upgrade notes about data loss.

PR #547, opened 2026-05-14
feat: TF service mesh outputs

MichaelThamm:
## Issue

Add TF outputs required for service mesh intergrations in products


adhityaravi (approved):
LGTM

CI Breaking - probably not because of this change

PR #565, opened 2026-06-23
fix: disable all analytics features in the config when requested

### LINKED ISSUE #504: Despite setting reporting_enabled to false in the charm config, Grafana analytics show as enabled in settings
### Bug Description

Despite setting reporting_enabled to false in the charm config, Grafana analytics show as enabled in Settings. 

See:

Image

### To Reproduce

1. `juju deploy grafana`
2. `juju config grafana reporting_enabled=false`
3. Navigate to the Settings page in Grafana and observe that analyitcs show as enabled.

### Environment

Juju 3.6.14, latest Grafana HA with 3 units.

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

gycsaba96:
## Issue

Closes #504

## Solution

Some extra (possibly new) options are now set to false in Grafana's analytics settings when `reporting_enabled` is set to false in the charm config.

**Side note:** the charm config name `reporting_enabled` is confusing, because Grafana has the same config option, but the charm actually controls a wider set of options. I propose to rename it to something more in-line with the functionality (e.g., `analytics_features_enabled`). I created an issue about it: #566.

## Context

https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#analytics

## Testing Instructions

1. Pack and deploy the charm:
```
charmcraft pack
juju deploy ./grafana-k8s_ubuntu@26.04-amd64.charm --trust --resource grafana-image=docker.io/ubuntu/grafana@sha256:1af2d3519cc3515708f855909a3a5329404079fd50d1cf6b69a7de40067447c4
```
2. Disable the analytics features:
```
juju config grafana-k8s reporting_enabled=false
```
3. Observe that analytics section has `false` values only in _Administration > General > Settings_.


src/grafana_config.py
```diff
@@ -214,6 +214,10 @@ def _generate_analytics_config(self) -> str:
         config_ini = configparser.ConfigParser()
         # Ref: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#analytics
         config_ini["analytics"] = {
+            "enabled": "false",
```
lucabello: This setting completely disables internal analytics, which is separate; as you also suggested, maybe we should rework this into a new config option called `analytics_features_enabled` or something?

gycsaba96: Based on the [description of the config](https://github.com/canonical/grafana-k8s-operator/blob/859486d5cb6d94d170b32f58e57443825b01f70f/charmcraft.yaml#L272), the original intent is to disable more than just that single option. This is also supported by the [current code](https://github.com/canonical/grafana-k8s-operator/blob/859486d5cb6d94d170b32f58e57443825b01f70f/src/grafana_config.py#L216) in the main branch. Finally, the screenshot in the bug report already has the `reporting_enabled` _Grafana_ config set to false.

But I fully agree that the name is misleading. :)

If we don't want to put them under the same umbrella, why not have a different charm config for each Grafana config? E.g., separately configuring update checks.

lucabello: The options you're disabling are valid, but I think `enabled=false` is the only one that's not, as that's **internal** analytics, not **external**; I think it'd be fine to remove setting `enabled=false` and just keep the other options, all operated by the same `reporting_enabled` config flag; what do you think?

gycsaba96: Sounds good, I will make the change.

gycsaba96: Done.

lucabello (approved):
LGTM, let's wait for CI to pass and merge :)

PR #567, opened 2026-06-23
feat: support Juju secrets for sensitive custom config fields

### LINKED ISSUE #507: SMTP charm config options store secrets in plain text
### Enhancement Proposal

The recently added custom config options, specifically SMTP support, store secrets in plain text in the charm config. Instead secrets should either:
1. Be added as secret-id references within the custom config
2. Added as separate charm config options that take Juju secrets

michaeldmitry:
Fixes #507 

Sensitive data in custom config provided by juju config can now be referenced from juju secrets using a `secret://<secret-id>/<key>` URL in place of the plaintext value


lucabello (approved):
`secrets_helper` seems like something we should (in the future) abstract either into a `charmlib` or some python package; can you create an issue to remind us to look into that?

The rest LGTM!

michaeldmitry:
> `secrets_helper` seems like something we should (in the future) abstract either into a `charmlib` or some python package; can you create an issue to remind us to look into that?
> 
> The rest LGTM!

I believe this is tracked in https://github.com/canonical/opentelemetry-collector-integrator-operator/issues/4

PR #570, opened 2026-06-25
fix(lib): only update dashboard databag when templates change

lucabello:
The grafana_dashboard library was generating a new UUID every time `_upset_dashboards_on_relation()` was called, even when the dashboard templates hadn't changed. This caused continuous databag updates and unnecessary relation-changed events on the consumer side.

The fix compares the new templates with the existing templates in the relation databag before updating. If they're identical, no update is made and the UUID remains stable.

This change affects both `GrafanaDashboardProvider` and `GrafanaDashboardAggregator` classes.

Fixes: canonical/opentelemetry-collector-operator#331

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -217,7 +217,7 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 49
+LIBPATCH = 50
 
 PYDEPS = ["cosl >= 0.0.50"]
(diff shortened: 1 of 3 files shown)


lucabello:
Manually tested the fix end-to-end:

1. Deployed patched otelcol and grafana-k8s from this branch to a test model
2. Established `grafana-dashboard` relation
3. Recorded initial UUID: `1ff96540-9a5a-4646-99c9-9d291af9d28f`
4. Fired 4 `update-status` events on otelcol via `juju exec`
5. UUID remained stable - no change after all events
6. Grafana status log confirmed no `grafana-dashboard-relation-changed` events fired after initial setup

Fix works as expected.

michaeldmitry (approved):

PR #578, opened 2026-07-07
fix: `grafana_dashboard` flaplint issues

sinapah:
## Issue

Based on https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/313

## Solution

```
Line 1372 (GrafanaDashboardProvider._upset_dashboards_on_relation) and Line 1870 (GrafanaDashboardAggregator._update_remote_grafana) use uuid.uuid4(), which generates a new ID on every execution and forces an infinite Juju hook loop. Replace str(uuid.uuid4()) with: `hashlib.sha256(json.dumps(new_templates, sort_keys=True).encode()).hexdigest()[:16]`
```
```
Line 1888 (GrafanaDashboardAggregator.remove_dashboards) lacks an equality guard. Extract the templates first, then hash them by replacing str(uuid.uuid4()) with: remaining_templates = type_convert_stored(...) and then hashing via hashlib.sha256(json.dumps(remaining_templates, sort_keys=True).encode()).hexdigest()[:16]
```
```
Line 1736 (GrafanaDashboardConsumer.set_peer_data) calls json.dumps(data) without sorting dict keys, causing arbitrary databag updates depending on Python's internal dictionary ordering. Force key sorting by changing it to: `json.dumps(data, sort_keys=True)`
```
```
Line 2137 (CosTool.validate_alert_rules) triggers a lint error due to a nondeterministic value written to a temporary file. Since this is an intentional, ephemeral file operation, suppress the linter warning by adding a comment on that line: `# databag-order: ignore`
```

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -184,7 +184,7 @@ import platform
 import re
 import subprocess
 import tempfile
-import uuid
+
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 import yaml
@@ -1373,7 +1373,7 @@ class GrafanaDashboardProvider(Object):
             "templates": new_templates,
             "uuid": hashlib.shake_128(
                 json.dumps(new_templates, sort_keys=True).encode()
-            ).digest(16).hex(),
+            ).digest(8).hex(),
         }
 
         relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
@@ -1875,7 +1875,7 @@ class GrafanaDashboardAggregator(Object):
                     "templates": new_templates,
                     "uuid": hashlib.shake_128(
                         json.dumps(new_templates, sort_keys=True).encode()
-                    ).digest(16).hex(),
+                    ).digest(8).hex(),
                 }
                 grafana_relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
 
@@ -1896,7 +1896,7 @@ class GrafanaDashboardAggregator(Object):
             "templates": remaining_templates,
             "uuid": hashlib.shake_128(
                 json.dumps(remaining_templates, sort_keys=True).encode()
-            ).digest(16).hex(),
+            ).digest(8).hex(),
         }
 
         if self._charm.unit.is_leader():
@@ -2141,8 +2141,8 @@ class CosTool:
             #       - alert: OtherAlert
             #         expr: up
             transformed_rules = {"groups": []}  # type: ignore
-            for rule in rules["groups"]:
-                transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
+            for i, rule in enumerate(rules["groups"]):
+                transformed = {"name": f"group_{i}", "rules": [rule]}
                 transformed_rules["groups"].append(transformed)
 
             rule_path.write_text(yaml.safe_dump(transformed_rules, sort_keys=True)) # databag-order: ignore
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -392,6 +392,13 @@ REACTIVE_CONVERTER = {  # type: ignore
 }
 
 
+def _data_hash(data: Any) -> str:
+    """Deterministic hash of a template dict for use as a stable relation data key."""
+    return hashlib.shake_128(
+        json.dumps(data, sort_keys=True).encode()
+    ).digest(8).hex()
+
+
 class RelationNotFoundError(Exception):
     """Raised if there is no relation with the given name."""
 
@@ -1371,9 +1378,7 @@ class GrafanaDashboardProvider(Object):
         # templates haven't changed, avoiding spurious relation-changed events.
         stored_data = {
             "templates": new_templates,
-            "uuid": hashlib.shake_128(
-                json.dumps(new_templates, sort_keys=True).encode()
-            ).digest(8).hex(),
+            "uuid": _data_hash(new_templates),
         }
 
         relation.data[self._charm.app]["dashboards"] = json.dumps(stored_data)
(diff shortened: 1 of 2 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -2134,7 +2145,7 @@ def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
                 transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
                 transformed_rules["groups"].append(transformed)
 
-            rule_path.write_text(yaml.dump(transformed_rules))
+            rule_path.write_text(yaml.dump(transformed_rules))  # databag-order: ignore
```
sinapah: `flaplint`'s README says:
>  It reads your charm's source code and flags every place a value that has no stable byte-order — a set, a glob, a uuid4(), … — reaches a churn-sensitive write: a relation databag, an on-disk file, a pebble plan, or a content-hash change-detector.

That's why I think this is being picked up, even though it only applies to a temp file. Just adding this ignore to silence the call.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1864,10 +1868,14 @@ def _update_remote_grafana(self, _: Optional[RelationEvent] = None) -> None:
...
                     continue  # No change in templates, don't update the databag
 
-                # It's still ridiculous to add a UUID here, but needed
+                # It's still ridiculous to add a UUID here, but needed.
+                # Use a deterministic hash of the templates so the value is stable when
+                # templates haven't changed, avoiding spurious relation-changed events.
                 stored_data = {
                     "templates": new_templates,
-                    "uuid": str(uuid.uuid4()),
+                    "uuid": hashlib.sha256(
```
sinapah: JUst a small note there that getting the first 16 chars of the hash is just arbitrary. I don't think we'll need to rely on the full 64 chars to determine whether the hash needs to change. This keeps the databag more neat and easier to look at.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1366,10 +1366,14 @@ def _upset_dashboards_on_relation(self, relation: Relation) -> None:
...
-        # pseudo-random value, this never makes it across 'juju set-state'
+        # pseudo-random value, this never makes it across 'juju set-state'.
+        # Use a deterministic hash of the templates so the value is stable when
+        # templates haven't changed, avoiding spurious relation-changed events.
         stored_data = {
             "templates": new_templates,
-            "uuid": str(uuid.uuid4()),
+            "uuid": hashlib.sha256(
+                json.dumps(new_templates, sort_keys=True).encode()
+            ).hexdigest()[:16],
```
sinapah: Setting the change from UUID to hash aside, I'm confused about this part of the code in the first place. The original comment seems to suggest that we add a UUID to trigger a relation changed event, but just a few lines above we have a `templates` equality guard. We only get here if the guard determines that the templates have changed, at which point a `relation-changed` event should be guaranteed (right?). Why have a UUID at all?

Maybe @sed-i has some context from before?

sed-i: This part of the code is from @rbarry82 times. `set-state` isn't about relation data but controller storage:

https://github.com/canonical/grafana-k8s-operator/blob/ab3c22560505ee613bf7fa415cf6fa6f02b58619/src/charm.py#L684

...which should be [deprecated](https://github.com/canonical/operator/pull/882) anyway.

I think we can keep your change for now.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: Shall we use a hash function with custom output length?

```python
>>> import hashlib
>>> hashlib.shake_128(b"data").digest(20).hex()
'6d5d29b8058bdf6517f2487f8dc537ab63749700'
>>> hashlib.shake_128(b"data").digest(20).hex()
'6d5d29b8058bdf6517f2487f8dc537ab63749700'
>>> hashlib.shake_128(b"data").digest(20).hex()
'6d5d29b8058bdf6517f2487f8dc537ab63749700'
```

Could also define elsewhere
```python
>>> def hasher(data):
...     return hashlib.shake_128(data).digest(16).hex()
...     
>>> hasher(b"data")
'6d5d29b8058bdf6517f2487f8dc537ab'
>>> hasher(b"data")
'6d5d29b8058bdf6517f2487f8dc537ab'
```

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sed-i: - Any reason not to use `safe_dump` over `dump`?
- `safe_dump` takes a `sort_keys` arg which could obviate the ignore.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -2130,11 +2141,11 @@ def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
...
             #         expr: up
             transformed_rules = {"groups": []}  # type: ignore
-            for rule in rules["groups"]:
-                transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
+            for i, rule in enumerate(rules["groups"]):
+                transformed = {"name": f"group_{i}", "rules": [rule]}
                 transformed_rules["groups"].append(transformed)
 
-            rule_path.write_text(yaml.dump(transformed_rules))
+            rule_path.write_text(yaml.safe_dump(transformed_rules, sort_keys=True)) # databag-order: ignore
```
sed-i, line 2149: So even with sort_keys we still need the ignore? Check with @michaeldmitry if it's a flaplint issue?

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1366,10 +1366,14 @@ def _upset_dashboards_on_relation(self, relation: Relation) -> None:
...
-        # pseudo-random value, this never makes it across 'juju set-state'
+        # pseudo-random value, this never makes it across 'juju set-state'.
+        # Use a deterministic hash of the templates so the value is stable when
+        # templates haven't changed, avoiding spurious relation-changed events.
         stored_data = {
             "templates": new_templates,
-            "uuid": str(uuid.uuid4()),
+            "uuid": hashlib.shake_128(
+                json.dumps(new_templates, sort_keys=True).encode()
+            ).digest(8).hex(),
```
sed-i: `shake128(...).digest().hex()` is repeated a couple of times, seems like a good candidate for function extraction?

sed-i (approved):
Nice! A couple of non blocking comments.

lib/charms/grafana_k8s/v0/grafana_dashboard.py
sinapah: Yes, extracted.

PR #587, opened 2026-08-13
feat: source cosTool from cosl

sinapah:
## Issue

Similar to: https://github.com/canonical/prometheus-k8s-operator/pull/846

## Solution

Also bump `cosl` to latest and also `charmcraft fetch-lib`.

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -180,16 +180,12 @@ import json
 import logging
 import lzma
 import os
-import platform
 import re
-import subprocess
-import tempfile
 
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple
-import yaml
-from cosl import DashboardPath40UID, LZMABase64
-from cosl.types import type_convert_stored
+from typing import Any, Callable, Dict, List, Optional
+from cosl import CosTool, DashboardPath40UID, LZMABase64
+from cosl.types import QueryType, type_convert_stored
 from ops.charm import (
     CharmBase,
     HookEvent,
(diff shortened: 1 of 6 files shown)


lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -1447,7 +1448,11 @@ def __init__(
         super().__init__(charm, relation_name)
         self._charm = charm
         self._relation_name = relation_name
-        self._transformer = CosTool(self._charm)
+        # A default query type is required here: cosl's `CosTool` only accepts the query
+        # type via the per-method `query_type` kwarg, but `_modify_panel` passes it
+        # positionally (per-panel, promql or logql), which the `ensure_querytype` guard
+        # does not see. "promql" is just a placeholder to satisfy that guard.
+        self._transformer = CosTool("promql")
```
sed-i, line 1455: The comment does say `promql` is a placeholder, but just making sure it's ok to set it here like that for dashboard panels whose datasource is loki rather than prom/mimir.

sinapah: Thanks for the careful review — it is safe. The constructor default is never actually used for the transform: in `_modify_panel` the query type is always resolved per-panel from the datasource and passed explicitly to `inject_label_matchers` (`${prometheusds}` -> "promql", `${lokids}` -> "logql"), and cosl's implementation does `query_type = query_type or self.query_type`, so the explicit per-call value always takes precedence over the default. Panels without a recognized datasource are skipped before we reach the transform, so every invocation carries an explicit promql or logql. The `"promql"` default is therefore only a placeholder to satisfy cosl's `ensure_querytype` guard, which requires a default query type (or a `query_type` kwarg) at the call site — and since `_modify_panel` passes it positionally, the guard would not see it without this default. The loki path is covered by `test_consumer_adds_labels_for_loki`, which renders a Loki panel and asserts the correct logql output.

